PR #1, opened 2018-10-23
2 functions to store the path of the uploaded MP3 files into the RT db

ahmad-elbatanouni:


functions/index.js
```diff
@@ -0,0 +1,72 @@
+
+const functions = require('firebase-functions');
+var admin = require("firebase-admin");
+
+
+admin.initializeApp({
+    credential: admin.credential.applicationDefault(),
+    databaseURL: "https://audio-seva-team-test.firebaseio.com",
```
anantakrishna: Don't hard-code any URL, use https://firebase.google.com/docs/functions/config-env

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+    storageBucket: 'audio-seva-team-test.appspot.com'
+});
+
+
+exports.on_mp3_upload = functions.storage.object().onFinalize( object => {
```
anantakrishna: Name functions according to the action, not the trigger. I think it might be better. As later we can add another action to the same trigger as a different function.

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+
+exports.currently_stored_mp3s_to_database = functions.https.onRequest((req, res) => {
+    
+    const bucket = admin.storage().bucket();
+    const db = admin.database();
+    bucket.getFiles().then(files => {        
+        files.forEach(innerFilesObject => {
+            innerFilesObject.forEach(file => {                
+                if(file.name.indexOf('.mp3') > 0) {
```
anantakrishna: Avoid duplicating the code here, DRY (don't repeat yourself).

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+    return Promise.resolve(1);
+});
+
+
+exports.currently_stored_mp3s_to_database = functions.https.onRequest((req, res) => {
+    
+    const bucket = admin.storage().bucket();
+    const db = admin.database();
+    bucket.getFiles().then(files => {
```
anantakrishna: Do you check for the storage path here? Only /mp3/ folder should be scanned.

functions/index.js
ahmad-elbatanouni: ok

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+    return Promise.resolve(1);
+});
+
+
+exports.currently_stored_mp3s_to_database = functions.https.onRequest((req, res) => {
+    
+    const bucket = admin.storage().bucket();
+    const db = admin.database();
+    bucket.getFiles().then(files => {        
+        files.forEach(innerFilesObject => {
```
anantakrishna: I don't understand what is `innerFilesObject`.

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+            innerFilesObject.forEach(file => {
```
anantakrishna: Please also check that the file is not yet present in the DB.

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+                if(file.name.indexOf('.mp3') > 0) {
+                    const parts = file.name.split('/');
+                    
+                    const list = parts[1];
+                    const mp3 = parts[2];
+                    let file_name = mp3.slice(0, -4);
+                    
+                    file_name = file_name.replace(/[.#$[\]]/g, '_');
+
+                    let ref = db.ref("/sqr/files/" + list + "/" + file_name);
```
anantakrishna: Doesn't JS allows here to use string interpolation instead of concatenating strings? I just don't remember.

functions/index.js
```diff
@@ -0,0 +1,72 @@
...
+        const db = admin.database();
+        let ref = db.ref("/sqr/files/" + list + "/" + file_name);
+        ref.set({status: "Spare"});
+    }
+
+    return Promise.resolve(1);
+});
+
+
+exports.currently_stored_mp3s_to_database = functions.https.onRequest((req, res) => {
```
anantakrishna:
```suggestion
exports.importMP3IntoSQR = functions.https.onRequest((req, res) => {
```

functions/.directory
```diff
@@ -0,0 +1,3 @@
+[Dolphin]
```
anantakrishna: Is this file necessary for the development? Is it an IDE file?

anantakrishna:
When I was adding the functions using Furebase CLI, it also changed the `firebase.json` file in the root, which is not included into your commit. Is it correct?

functions/index.js
ahmad-elbatanouni: checked

functions/index.js
ahmad-elbatanouni: checked

functions/index.js
ahmad-elbatanouni: checked

functions/index.js
ahmad-elbatanouni: checked

functions/index.js
ahmad-elbatanouni: yes it's allowed... replaced the old way of concatenation to string interpolation

functions/index.js
ahmad-elbatanouni: checked

functions/.directory
ahmad-elbatanouni: no it's related to the OS, I'll add it to .gitignore

functions/index.js
ahmad-elbatanouni: it's returning an array of arrays of File objects
screenshot_20181023_203208
screenshot_20181023_203222

ahmad-elbatanouni:
ok, I'll remove the nested folder and upload the files directly under the "functions" folder
Regarding the **firease.json** file, the one I have has nothing but a lint "test" to be run before deploying to firebase... do I need to upload it to ?

functions/index.js
```diff
@@ -0,0 +1,66 @@
...
+    bucket.getFiles().then(files => {
+        console.log('Outer')
+        console.log(files.toString());
+
+        files.forEach(innerFilesObject => {
+            console.log('Inner')
+            console.log(innerFilesObject.toString());
+
+            innerFilesObject.forEach(file => {
+                if(file.name.indexOf("mp3/") > -1 && file.name.indexOf('.mp3') > 0)
```
anantakrishna: Isn't this condition still duplicated?

functions/index.js
```diff
@@ -0,0 +1,66 @@
+
+const functions = require('firebase-functions');
+var admin = require("firebase-admin");
+
+
+admin.initializeApp({
+    credential: admin.credential.applicationDefault(),
+    databaseURL: functions.config().audioseva.database_url,
```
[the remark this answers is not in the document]
ahmad-elbatanouni: the credentials will be set automatically, 
but the **database** and the **storage bucket** URLs will have to be set manually as in the example given here:
https://firebase.google.com/docs/functions/config-env

anantakrishna:
Ok, thanks for the clarification. Document necessary configuration in the
readme file in the root.

functions/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+        return 1;
+    }).catch(err => console.log(err));
+}
+
+
+exports.importMP3IntoSQR = functions.storage.object().onFinalize( object => {
+    const filePath = object.name;
+    
+    if(filePath.indexOf("mp3/") > -1 && filePath.indexOf(".mp3") > 0)
```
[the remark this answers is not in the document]
ahmad-elbatanouni: checked

functions/index.js
ahmad-elbatanouni: removed redundant IF

PR #2, opened 2018-10-23
Update sqr filter to use Firebase as source of data for retrieving files

DadImScared:


frontend/src/views/SQRAllotment.vue
```diff
@@ -136,23 +138,33 @@ export default {
...
-            }
-          })
-          .then(response => {
-            this.files = response.body;
-          });
+        const fetchData = await fb
+          .database()
+          .ref(`sqr/files/${this.filter.list}`)
+          .once("value");
+        const files = fetchData.val();
```
anantakrishna: Here you actually pull all the records and filter in the browser. Isn't it possible to [filter server-side](https://firebase.google.com/docs/database/admin/retrieve-data#section-queries). Maybe for this sake we'd need to create [indexes in db](https://firebase.google.com/docs/database/security/#section-defining-indexes).

frontend/src/views/SQRAllotment.vue
```diff
@@ -136,23 +138,33 @@ export default {
...
-            params: {
-              path: "sqr/files",
-              list: this.filter.list,
-              language: this.filter.language
-            }
-          })
-          .then(response => {
-            this.files = response.body;
-          });
+        const fetchData = await fb
```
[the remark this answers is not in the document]
DadImScared: I did not know that library exists. I will switch over to it.

frontend/src/views/SQRAllotment.vue
DadImScared: I am having trouble figuring out the best way to filter server-side since I need to filter by two fields but firebase only allows one field. Do you have any suggestions?

anantakrishna: Seems you're right. According to [this answer](https://stackoverflow.com/a/26701282) I propose that we rethink the data structure and build our custom indexes in the DB. Really, NoSQL is different, and the data cannot be normalized as in SQL. So we'd have to push data into several nodes and/or move the data between nodes on some changes (allotments etc).

Will create a task in Trello for this, for now we can keep this as it is to close this task.

anantakrishna: BTW, there is a library Querybase, however they claim that Firestore is more featured with querying… 
image

Do you think we may look into Firestore? I didn't want to utilize it because of it's Beta stage and the pricing differences.

DadImScared: Firestore looks promising for filtering data how we'd like. I am also concerned with the pricing and Beta stage but do not know the full scale of the project so I can not comment about it.

anantakrishna:
But you know, I have a thought. Mainly we need to filter by status, and
secondly by language. So let's keep Realtime Database for now and filter by
status on the server and by language on the client.

anantakrishna:
Are all the issues fixed? Is this ready for the next review?

DadImScared:
It is now, I forgot to remove some unused code.

anantakrishna:
Can you please resolve the conflicts? I've merged another pull request before yours, so that is why conflicting. Maybe we can turn off `package-lock` file? It will cause conflicts always.

PR #3, opened 2018-10-23
Setup jest

DadImScared:


DadImScared:
It added it when I ran `vue add @vue/unit-jest`

PR #7, opened 2018-10-26
Sqr allotment

DadImScared:


DadImScared:
I had trouble accessing `this.$firebaseRefs` based on the vuefire documentation so the base sdk is used to post data.

PR #14, opened 2018-11-11
Transfer SQR data from the spreadsheet into Database

ahmad-elbatanouni:
The spreadsheet url is hardcoded cause I didn't where it should come from .. will it be an environment variable or should it be read from a path at the database.

### DIFF
--- a/functions/src/helpers.ts
+++ b/functions/src/helpers.ts
@@ -7,6 +7,30 @@ apiKey.apiKey = sendInBlueSecretKey;
 
 
 
+/////////////////////////////////////////
+//  Groups an array by a particular index
+//
+//  example cars = [ { make: 'audi', year: '2012' }, 
+//                   { make: 'ford', year: '2015' },
+//                   { make: 'audi', year: '2018' } 
+//                 ]
+//
+//  groupListBy( cars, "make" )
+//      output: {   audi: [ { make: 'audi', year: '2012' },
+//                          { make: 'audi', year: '2018' } ],
+//
+//                  ford: [{ make: 'ford', year: '2015' }]
+//              }
+/////////////////////////////////////////
+export const groupListBy = (array, index) => {
+    return array.reduce((r, a) => {
+        r[a[index]] = r[a[index]] || [];
+        r[a[index]].push(a);
+        return r;
+    }, Object.create(null));
+};
+
+
 export const checkValidMP3 = filePath => (filePath.startsWith("mp3/") && filePath.endsWith(".mp3"))
 
 export const createMP3DBRef = (filePath, _module) => {
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -224,6 +224,26 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
 //      2. Looks for two sheets --> Allotments & Submissions
 //      3. Loads their data into the equivalent Firebase database paths
 /////////////////////////////////////////////////
+
+
+// Helper Function
+//      splits an array into a bunch of arrays 
+//      GROUPED BY a 
+//              composite key ( 2nd parameter: values )
+
+let groupByMulti = (list, values, context) => {
+    if (!values.length) {
+      return list;
+    }
+    var byFirst = lodash.groupBy(list, values[0], context),
+        rest    = values.slice(1);
+    for (var prop in byFirst) {
+      byFirst[prop] = groupByMulti(byFirst[prop], rest, context);
+    }
+    return byFirst;
+}
+
+
 export const importSpreadSheetData = functions.https.onRequest( async (req, res) => {
     const auth = await google.auth.getClient({
         scopes: ["https://www.googleapis.com/auth/spreadsheets"]
@@ -253,6 +273,24 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
     const submissions = await getSubmissions();
     submissions.forEach(row => {
         let serial = row['submissionserial'];
+
+        const regex = /(.*?)–(.*):(.*)—(.*)/g;
+        let soundissuesMatch = regex.exec(row['soundissues']);
+        let unwantedpartsMatch = regex.exec(row['unwantedparts']);
+        
+        let soundissues = {
+            beginning: soundissuesMatch[1],
+            ending: soundissuesMatch[2],
+            type: soundissuesMatch[3],
+            description: soundissuesMatch[4]
+        },
+        unwantedparts = {
+            beginning: unwantedpartsMatch[1],
+            ending: unwantedpartsMatch[2],
+            type: unwantedpartsMatch[3],
+            description: unwantedpartsMatch[4]
+        };
+
         let submission = {
             author: {
                 name: row['name'],
@@ -263,9 +301,9 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
             completed: row['completed'],
             created: row['created'],
             comments: row['comments'],
-            soundissues: row['soundissues'],
+            soundissues,
             soundqualityrating: row['soundqualityrating'],
-            unwantedparts: row['unwantedparts'],
+            unwantedparts,
             duration: {
                 beginning: new Date(row['beginning']).getTime() / 1000,
                 ending:  new Date(row['ending']).getTime() / 1000,
@@ -281,39 +319,26 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
     const allotments = await getAllotments();
    
         
-    // Grouping all the files allotted on one day under a single `Allotment Node` in the db
-
-    // 1 Group by ASSIGNEEs
-    let Assignees = lodash.groupBy(allotments, "devotee");
-
-    // 2 Group by ASSIGNEEs/Dates
-    let AssigneesDates = {};
-    for (let i in Assignees)
-        AssigneesDates[i] = lodash.groupBy(Assignees[i], "dategiven");
-
-    // 3 Group by ASSIGNEEs/DATEs/LISTs
-    let AssigneesDatesLists = {};
-    for (let assignee in AssigneesDates) {
-        AssigneesDatesLists[assignee] = {};
-        for (let date in AssigneesDates[assignee])
-            AssigneesDatesLists[assignee][date] = lodash.groupBy(AssigneesDates[assignee][date], "list")
-    }
+    // Group all the files allotted on one day under a single `Allotment Node` in the db
+    // Group by ASSIGNEEs/DATEs/LISTs
+    
+    let groupedAllotments = groupByMulti(allotments, ['devotee', 'dategiven', 'list'], {});
 
     // Adding the allotments
-    for (let assignee in AssigneesDatesLists) {
-        for (let date in AssigneesDatesLists[assignee]) {
+    for (let assignee in groupedAllotments) {
+        for (let date in groupedAllotments[assignee]) {
             let dayFiles = [];
-            for (let list in AssigneesDatesLists[assignee][date] ) {
+            for (let list in groupedAllotments[assignee][date] ) {
                 
                 // Collecting all of the files on a list under a single day in an array
-                AssigneesDatesLists[assignee][date][list].forEach(item => {
+                groupedAllotments[assignee][date][list].forEach(item => {
                     dayFiles.push(item['filename']);
-                })
+                });
                 
                 let allotment = {
                     assignee: {
                         name: assignee,
-                        emailAddress: Assignees[assignee][0]['email'],
+                        emailAddress: groupedAllotments[assignee][date][list][0]['email'],
                     },
                     files: dayFiles,
                     list,
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -33,23 +33,35 @@ export const importFilesFromStorage = functions.storage.object()
 
 export const updateFilesOnNewAllotment = functions.database.ref('/sqr/allotments/{allotment_id}')
 .onCreate((snapshot, context) => {
-    const original = snapshot.val();
+    const allotment = snapshot.val();
     let newDocKey = snapshot.key;
-    original.files.forEach(async file => {
-        let file_ref = db.ref(`/sqr/files/${original.list}/${file}`);
+    allotment.files.forEach(async file => {
+        let file_ref = db.ref(`/files/${allotment.list}/${file}/soundQualityReporting`);
         const snapshot = await file_ref.child("status").once('value');
         if(snapshot.exists()) {
-            file_ref.update(
-            {
+            file_ref.update({
                 status: 'Given',
                 allotment: {
                     timestampGiven: new Date().getTime(),
                     timestampDone: null,
-                    devotee: original.devotee
+                    asignee: allotment.asignee
                 }
-            }, err => {
-                if (!err)
-                    db.ref(`/sqr/allotments/${newDocKey}`).update({ filesAlloted: true });
+            }, (err) => {
+                if (!err) { // if Successful FILE Update, update the ALLOTMENT accordingly
+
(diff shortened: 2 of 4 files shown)


functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+//      1. Parses a google spreadsheet
+//      2. Looks for two sheets --> Allotments & Submissions
+//      3. Loads their data into the equivalent Firebase database paths
+/////////////////////////////////////////////////
+export const importSpreadSheetData = functions.https.onRequest( async (req, res) => {
+    const auth = await google.auth.getClient({
+        scopes: ["https://www.googleapis.com/auth/spreadsheets"]
+    });
+    
+    const spreadsheetUrl = 'https://docs.google.com/spreadsheets/d/1e0rYguPua_0Rtd--uNLKkGotNw-CPM44TlBS5hthdfQ/edit#gid=477553865';
```
anantakrishna: Spreadsheet ID should be taken from `config.sqr.spreadsheetId`.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+    // extracting the spreadsheet ID from the url
+    const spreadsheetId = new RegExp("/spreadsheets/d/([a-zA-Z0-9-_]+)").exec(spreadsheetUrl)[1];
```
anantakrishna: No need parsing URL, just get spreadsheet Id from the config.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+    const spreadsheet = new GoogleSpreadsheet(spreadsheetId);
+
+    const token = await auth.getAccessToken();
+    spreadsheet.setAuthToken(token);
+
+    spreadsheet.getInfo((err, data) => {
```
anantakrishna: Isn't it possible to promisify this call and even better to use `async/await`? Code is hardly readable.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+        if (!err) {
+            data.worksheets.forEach(worksheet => {
```
anantakrishna: Instead of `forEach` and `if` better to directly get two particular worksheets if possible.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                if(worksheet.title === "Submissions")
+                    worksheet.getRows({ }, (err, rows) => {
+                        rows.forEach(row => {
+                            let documentID = row['submissionserial'];
+                            db.ref(`/sqr/submissions/s${documentID}`)
```
anantakrishna: Is there a typo — `s` before document Id?

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                                .set({
+                                    devotee: {
+                                        name: row['devotee'],
+                                        emailAddress: row['email'],
+                                    },
```
anantakrishna: Here is something wrong. Firstly, still `devotee` is here. Secondly, column names are not matching the [Submissions sheet](https://docs.google.com/spreadsheets/d/1e0rYguPua_0Rtd--uNLKkGotNw-CPM44TlBS5hthdfQ/edit#gid=729796178)

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                                    fileName: row['audiofilename'],
+                                    changed: row['changed'],
+                                    completed: row['completed'],
+                                    created: row['created'],
```
anantakrishna: I'm really sure here also columns are not matching the sheet.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                                    comments: row['comments'],
+                                    soundissues: row['soundissues'],
+                                    soundqualityrating: row['soundqualityrating'],
+                                    unwantedparts: row['unwantedparts'],
+                                });
```
anantakrishna: Where is `beginning` and `ending`?

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                        })
+                    });
+
+                
+                if(worksheet.title === "Allotments")
+                    worksheet.getRows({}, (err, rows) => {
+                        rows.forEach(row => {
+                            db.ref(`/sqr/submissions`)
```
anantakrishna: Why `submissions`?

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                            .push({
+                                devotee: {
```
anantakrishna: Still `devotee`.

functions/src/modules/SQR.functions.ts
```diff
@@ -248,3 +251,73 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+                                    name: row['devotee'],
+                                    emailAddress: row['email'],
+                                },
+                                files: [ row['filename'] ],
+                                list: row['list'],
+                                comment: row['notes']
+                            });
```
anantakrishna: Where is the `timestamp` field?

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: ok

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: ok

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: ok

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: the **[API](https://www.npmjs.com/package/google-spreadsheet)**  has no such method

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: yes, sorry for that

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: 1. I worked on this before you told us about restructuring the data ( I was just waiting for your comments here before moving on with the restructures data )
2. Column names are converted by the Node library ( lower-cased, and then spaces are removed ) ... I inspected the object first before using the column titles.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: sorry, edited

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: It doesn't exist on the file
image

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: Didn't really know where they should be placed ... added them now under `duration`
image

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: image
image

ahmad-elbatanouni:
[SQR: Allotment processing](https://trello.com/c/hEX4y4zd/12-sqr-allotment-processing)

functions/src/modules/SQR.functions.ts
```diff
@@ -32,23 +33,35 @@ export const importFilesFromStorage = functions.storage.object()
...
     let newDocKey = snapshot.key;
-    original.files.forEach(async file => {
-        let file_ref = db.ref(`/sqr/files/${original.list}/${file}`);
+    allotment.files.forEach(async file => {
+        let file_ref = db.ref(`/files/${allotment.list}/${file}/soundQualityReporting`);
         const snapshot = await file_ref.child("status").once('value');
         if(snapshot.exists()) {
-            file_ref.update(
-            {
+            file_ref.update({
```
anantakrishna: Why do you check here for `status` to exist?

functions/src/modules/SQR.functions.ts
```diff
@@ -32,23 +33,35 @@ export const importFilesFromStorage = functions.storage.object()
...
                 status: 'Given',
                 allotment: {
                     timestampGiven: new Date().getTime(),
                     timestampDone: null,
-                    devotee: original.devotee
+                    asignee: allotment.asignee
+                }
+            }, (err) => {
+                if (!err) { // if Successful FILE Update, update the ALLOTMENT accordingly
+
```
anantakrishna: Isn't it possible to handle the error state using `async`?

functions/src/modules/SQR.functions.ts
```diff
@@ -32,23 +33,35 @@ export const importFilesFromStorage = functions.storage.object()
...
+                    // case 1 -- the allotmnet is read from the spreadsheet
+                    if (Object.keys(allotment).indexOf('sendNotificationEmail') > -1)
```
anantakrishna: I don't understand the condition here.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
+
+
+    data.worksheets.forEach(async worksheet => {
+        const getRows = promisify(worksheet.getRows);
+
+        if(worksheet.title === "Submissions") {
+            let rows = await getRows({});
+            rows.forEach(row => {
+                let documentID = row['submissionserial'];
+                let spreadsheetRow = {
```
anantakrishna: Here you are constructing a submission to be added to the database, not a spreadsheet row. Please rename.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
+                    author: {
+                        name: row['name'],
+                        emailAddress: row['emailAddress'],
+                    },
+                    fileName: row['audiofilename'],
+                    changed: row['changed'],
+                    completed: row['completed'],
+                    created: row['created'],
+                    comments: row['comments'],
+                    soundissues: row['soundissues'],
```
anantakrishna: I guess this needs parsing, as in the spreadsheet sound issues are formatted as text, but in the database it should be structured.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
+                    soundqualityrating: row['soundqualityrating'],
+                    unwantedparts: row['unwantedparts'],
+                    duration: {
+                        beginning: row['beginning'],
+                        ending: row['ending'],
+                    }
```
anantakrishna: Convert timings into seconds from the beginning of the file.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
+    let data = await getInfo();
+
+
+    data.worksheets.forEach(async worksheet => {
+        const getRows = promisify(worksheet.getRows);
+
+        if(worksheet.title === "Submissions") {
+            let rows = await getRows({});
+            rows.forEach(row => {
+                let documentID = row['submissionserial'];
```
anantakrishna: Why not to name it  `serial` for example? It's more appropriate and readable.

functions/src/helpers.ts
```diff
@@ -7,6 +7,30 @@ apiKey.apiKey = sendInBlueSecretKey;
...
+/////////////////////////////////////////
+export const groupListBy = (array, index) => {
+    return array.reduce((r, a) => {
+        r[a[index]] = r[a[index]] || [];
+        r[a[index]].push(a);
+        return r;
+    }, Object.create(null));
+};
+
+
```
anantakrishna: Are you sure there is no such a function in `lodash` package or any else? I'm against inventing the wheel.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
+        if(worksheet.title === "Allotments"){
+            let rows = await getRows({});
+
+            // Group all the files allotted on one day under a single `Allotment Node` in the db
+
+            // 1 Group by ASSIGNEEs
+            let assignees = helpers.groupListBy(rows, "devotee");
+
+            // 2 Group by ASSIGNEEs/Dates
+            let AssigneesDates = {};
```
anantakrishna: Local variables should be camel-cased.

functions/src/modules/SQR.functions.ts
```diff
@@ -265,57 +283,94 @@ export const importSpreadSheetData = functions.https.onRequest( async (req, res)
...
-                    });
-            });
+                        let allotment = {
+                            asignee: {
+                                name: assignee,
+                                emailAddress: assignees[assignee][0]['email'],
+                            },
+                            files: dayFiles,
+                            list,
+                            timestamp: date,
```
anantakrishna: Needs converting spreadsheet date into UNIX timestamp?

functions/src/modules/SQR.functions.ts
anantakrishna: Yes, I'm about the readability. You could create an index of the worksheets and then get the required worksheet by name.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: Once a new allotment is added manually I update it to include a `sendNotificationEmail` flag, unlike the ones read from the spreadsheet.

So if the allotment has this `sendNotificationEmail` **property** it means it's manually added and **requires an email notification**.
But if it doesn't, then no emails should be sent as you mentioned in the checklist [here](https://trello.com/c/bbuzNU2l/11-transfer-sqr-data-from-the-spreadsheet-into-database)

functions/src/helpers.ts
ahmad-elbatanouni: No, There is .. just thought it's useless to import the library for something silly like that.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: the problem here is that there's no consistent format, some row has `—` as a separator, some others have `-`.
I'll leave it like that for now until I get a feedback.

ahmad-elbatanouni:
`Sound Issues` field is still not parsed because I'm not sure how I should do it ... sometimes the separator between the beginning, end, desc, and type are `-` ,,, sometimes it's `—`.
I still need your opinion on this ... but it should take no time to parse it once we agree on a way to do it.

ahmad-elbatanouni:
I realize you hate nested loops, but I couldn't really think of any other way to iterate through the allotments grouped by the composite key.

anantakrishna:
> I realize you hate nested loops, but I couldn't really think of any other way to iterate through the allotments grouped by the composite key.

I dislike them. Anyhow, I will not insist here.

PR #17, opened 2018-11-13
Refactor sqr statistics to use new data structure

DadImScared:


DadImScared:
[SQR: Statistics page](https://trello.com/c/Jxw4lE2Z/19-sqr-statistics-page)

DadImScared:
Yes, I could not find a way to do it with firebase.

anantakrishna:
Perhaps we have to maintain statistics in the real-time using functions…
Anyhow, we'll refactor this later.

PR #21, opened 2018-11-18
Email Notifications Engine

ahmad-elbatanouni:
This is not the final version of course, This is just to help explaining my questions on trello, so excuse me if there were a lot of comments.


ahmad-elbatanouni:
[Email Notifications Engine](https://trello.com/c/mQPPsccd/21-email-notifications-engine)

functions/src/helpers.ts
```diff
@@ -5,7 +5,7 @@ const defaultClient = SibApiV3Sdk.ApiClient.instance;
 let apiKey = defaultClient.authentications['api-key'];
 apiKey.apiKey = sendInBlueSecretKey;
 
-
+const apiInstance = new SibApiV3Sdk.SMTPApi();
```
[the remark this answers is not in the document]
anantakrishna: Put all the notification sending stuff into a separate file `email.functions.ts`.

anantakrishna:
And by the way rework SQR function to make it not sending the email directly but rather pushing a notification into the database. But once again, move all the code that is related to this Email Engine into a separate file, don't mix things together.

functions/src/helpers.ts
```diff
@@ -29,3 +29,46 @@ export const sendEmail = (to, bcc, templateId, params) => {
...
+    return apiInstance.createSmtpTemplate(smtpTemplate).then(data => {
+      return data['id'];
+    },err => console.error(err)).catch(err => console.log(err));
+}
+
+
+export const getTemplateId = (templateName) => {
+    var opts = { 'templateStatus': true };
+    
+    return apiInstance.getSmtpTemplates(opts).then(function(data) {
```
anantakrishna: Can be `await/async` used here?

functions/src/modules/SQR.functions.ts
```diff
@@ -205,3 +206,87 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+    const os = require('os');
+    const fs = require('fs');
+
+    const filePath = object.name;
+    // slice(0, -5) to get rid of the trailing '.html' extension
+    const fileName = filePath.split('/')[2].slice(0, -5);
+    const tempLocalFile = path.join(os.tmpdir(), fileName);
+    
+    
+    if (!filePath.startsWith('email/templates'))
```
anantakrishna: That is why I was asking about a separate bucket. As this function will be fired on every other files upload, which is not good. Please comment in the card.

anantakrishna: That is why I was asking about a separate bucket. As this function will be fired on every other files upload, which is not good. Please comment in the card.

functions/src/modules/SQR.functions.ts
```diff
@@ -205,3 +206,87 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+    // slice(0, -5) to get rid of the trailing '.html' extension
+    const fileName = filePath.split('/')[2].slice(0, -5);
+    const tempLocalFile = path.join(os.tmpdir(), fileName);
+    
+    
+    if (!filePath.startsWith('email/templates')) 
+        return 1;
+
+    let emailRef = bucket.file(object.name);
+    await emailRef.download({ destination: tempLocalFile });
```
anantakrishna: Maybe we can use a public URL of the file and pass it to the SendInBlue API rather than downloading it? Just consider, I'm not insisting.

functions/src/helpers.ts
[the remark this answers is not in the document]
ahmad-elbatanouni: Should I give it the `.functions` extension ?
I think it's better if we keep this extension reserved only to the cloud functions files.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: I tried it at first but it gave me a permission denied error `SigningError: Permission iam.serviceAccounts.signBlob is required to perform this operation on service account`
and I kinda got lost trying to find an answer on how to do it without a service account.
If you find it important I can retry and search again.

functions/src/helpers.ts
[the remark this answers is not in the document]
anantakrishna: This file will contain cloud functions, isn't it?

functions/src/modules/SQR.functions.ts
anantakrishna: Let's keep it for now.

README.md
```diff
@@ -6,7 +6,7 @@ The main credentials, **databaseURL** & the **storageBucket** are automatically
 # each arg must have at least 2-part key (e.g foo.bar)
 $ firebase functions:config:set send_in_blue.key="sendInBlue secret Key"
 # Coordinator details
-$ firebase functions:config:set sqr.allotment.templateid=NUMBER
+$ firebase functions:config:set sqr.allotment.templateid='String | template name'
```
anantakrishna: This is not needed anymore as we will identify templates by name.

functions/src/modules/SQR.functions.ts
```diff
@@ -190,17 +191,17 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
-                {
+            db.ref(`/email/notifications`).push({
+                template: templateId,
+                to: coordinator.email_address,
+                bcc: [{ email: coordinator.email }],
+                params: {
                     submission,
                     fileData,
-                    devoteeAllotmentsSet: [],
+                    assigneeAllotmentsSet: [],
```
anantakrishna: `currentSet`

functions/src/modules/SQR.functions.ts
```diff
@@ -214,65 +215,71 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
-    const path = require('path');
-    const os = require('os');
-    const fs = require('fs');
-
-    const filePath = object.name;
-    // slice(0, -5) to get rid of the trailing '.html' extension
-    const fileName = filePath.split('/')[2].slice(0, -5);
-    const tempLocalFile = path.join(os.tmpdir(), fileName);
-    
+const updateEmailTemplates = async (filePath) => {
```
anantakrishna: You misunderstood the idea of moving all the email stuff into a separate file. This function and below — all should go to `email.functions.ts`. In `SQR` file you should only push to `/email/notifications`.

functions/src/modules/SQR.functions.ts
```diff
@@ -214,65 +215,71 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
     // 1. Get the template ID (sendInBlue ID)
-    //////////////////////////////////////////
-
     let templateNode = await db.ref('/email/templates').orderByKey()
-    .equalTo(fileName).once('value');
-
+                                .equalTo(fileName).once('value');
     let template = templateNode.val();
 
     if (template.exists()) {
```
anantakrishna: Again a long `if`. Refactor it to make it short.

functions/src/modules/SQR.functions.ts
```diff
@@ -214,65 +215,71 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+        await emailRef.download({ destination: tempLocalFile });
+
+        // Use the following to upate the template on SendInBlue
+        let htmlContent = fs.readFileSync(tempLocalFile, 'utf-8');
+        let { newTemplate, sender, bcc, subject, sample } = template;
+        let id;
+
+        if (newTemplate) {
+            id = await emailHelper.createTemplate(fileName, htmlContent, subject);
+            await db.ref(`/email/templates/${fileName}`).push({
```
anantakrishna: You should not push into this path. The template metadata should be created manually.

functions/src/modules/SQR.functions.ts
```diff
@@ -214,65 +215,71 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
-    } 
-    else {
-        let subject = 'WHERE CAN I GET subject FROM'; // mandatory
-        let id = await helpers.createTemplate(fileName, htmlContent, subject);
-        await db.ref(`/email/templates/${fileName}`).push({
-            id,
-            lastUpdated: new Date(),
-            version: 1
-            // Need some way to fill `sample`
+        db.ref(`/email/notifications`).push({
```
anantakrishna: Test email should be sent using a dedicated SendInBlue API call.

functions/src/modules/SQR.functions.ts
```diff
@@ -214,65 +215,71 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
+}
+
+export const updateTemplatesOnTemplateUpload = functions.storage.object()
+.onFinalize(object => { // called when either a NEW object is created, or when an object is overwritten
+    updateEmailTemplates(object.name);
     return 1;
 });
 
+export const updateTemplatesOnMetadataChange = functions.database.ref('/email/templates}')
+.onUpdate(async (change, context) => {
```
anantakrishna: Why only `onUpdate`? Or it is also fired on creating a node?

functions/src/modules/SQR.functions.ts
```diff
@@ -4,6 +4,7 @@ import * as  admin from 'firebase-admin';
 const bucket = admin.storage().bucket();
 const db = admin.database();
 import * as helpers from './../helpers';
+import * as emailHelper from './email.functions';
 
 let emailTemplates = {};
```
anantakrishna: The reason for this global variable is unclear. I cannot see any code adding to this object.

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const sendEmail = async (to, bcc, templateId, params) => {
+    let sendEmail = new SibApiV3Sdk.SendSmtpEmail();
+
+    sendEmail = {
+        to: [{ email: to }],
+        bcc,
+        templateId,
+        params
+    };
```
anantakrishna: It is a bad practice to name a variable as a function. And the variable should be a noun, not a verb. And you're overwriting it here. So three issues here.

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const createTemplate = async (templateName, html, subject) => {
+    let smtpTemplate = new SibApiV3Sdk.CreateSmtpTemplate();
+
+    smtpTemplate.templateName = templateName;
+    smtpTemplate.htmlContent = html;
+    smtpTemplate.subject = subject;
+    smtpTemplate.sender = {
+      email: functions.config().coordinator.email_address // ASSUMED Until a comment is made upon it
+    };
```
anantakrishna: Sender for a template is stored in the database.

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const getTemplateId = async (templateName) => {
+    var opts = { 'templateStatus': true };
+    
+    let result = await apiInstance.getSmtpTemplates(opts)
+    let { templates } = result;    
+    for (let i = 0; i < templates.length; i++) {
+        if (templates[i]['name'] === templateName)
+            return templates[i]['id'];
+    }
```
anantakrishna: Consider using any other means of iterating to make the code more readable.

functions/src/helpers.ts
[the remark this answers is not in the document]
ahmad-elbatanouni: No, it's just used to call sendInBlue API methods

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: I don't think we should trigger the function on **creating** a new node as the template wouldn't exist yet at the storage... 
A node is created and then on uploading the template that matches the `template` field in this node the function is automatically triggered.

README.md
ahmad-elbatanouni: it's still needed in sending mail to an assignee on a anew allotment...
it's not specified in the **SQR Allotments** data strcuture.

functions/src/modules/SQR.functions.ts
```diff
@@ -199,17 +194,17 @@ export const processSubmissions = functions.database.ref('/webforms/sqr/{submiss
...
             // Sending the notification Email Finally
-            helpers.sendEmail(
-                coordinator.email_address,
-                [{ email: coordinator.email }],
-                templateId,
-                {
+            db.ref(`/email/notifications`).push({
+                template: templateId,
+                to: coordinator.email_address,
+                bcc: [{ email: coordinator.email }],
```
anantakrishna: No need to add the coordinator to `bcc` here, as he is in `to` already.

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,100 @@
...
+const bucket = admin.storage().bucket();
+const db = admin.database();
+import * as helpers from './../helpers';
+
+const path = require('path');
+const os = require('os');
+const fs = require('fs');
+
+
+let emailTemplates = {};
```
anantakrishna: Repeating: this variable is never reassigned or pushed to. What is the purpose of it?

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,100 @@
...
+    const data = snapshot.val();
+    const templateName = snapshot.key;
+
+    let id;
+    if (Object.keys(emailTemplates).indexOf(templateName) > -1)
+        id = emailTemplates[templateName].id;
+    else 
+        id = await helpers.getTemplateId(templateName)
+
+    await helpers.sendEmail(data.to, data.bcc, id, data.params);
```
anantakrishna: `replyTo` to is missing. Please add. I have added it into the data structure.

functions/src/modules/email.functions.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    return snapshot.ref.update({ sent: true });
```
anantakrishna: I think it might be more useful to save the timestamp when the notification was actually sent. It would help debugging in the future. Of course, it should be named differently, like `sentTimestamp`.

And please test that this property is not filled to prevent double sending.

functions/src/index.ts
```diff
@@ -24,5 +24,6 @@ try {
 **********************************************/
 import * as SQR from './modules/SQR.functions';
 import * as ReportingSync from './modules/reporting.sync.functions';
+import * as email from './modules/email.functions';
```
anantakrishna: Let's be consistent. Let all the imports here be in a Initial Capital case.

functions/src/helpers.ts
```diff
@@ -1,30 +1,77 @@
...
+    let template = templates.filter(temp => temp['name'] === templateName)[0]
+
+    if (template === undefined)
+        return -1; // template not found
+    else
+        return templates['id'];
+}
+
+
+export const sendTestTemplate = async (templateId, emailTo) => {
```
anantakrishna: As far as I know, you may omit the `emailTo`, as SendInBlue will send to a predefined email list.

functions/src/helpers.ts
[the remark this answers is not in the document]
anantakrishna: No-no, all the email stuff should go to the `email.functions.ts` file. It is not needed anywhere except there.

functions/src/modules/email.functions.ts
ahmad-elbatanouni: It's used for creating a caching index ... in `sendNotificationEmail` 
```js
if (Object.keys(emailTemplates).indexOf(templateName) > -1)
        id = emailTemplates[templateName].id;
```

functions/src/modules/email.functions.ts
ahmad-elbatanouni: I'm not actually getting the point out of using a flag of any kind here... the Firebase function watches only **newly created** nodes .. I created it as you suggested now .. but I'm just asking

functions/src/modules/email.functions.ts
anantakrishna: Yes I see, but you're reading from this index here, and not writing anywhere. That's what confused me. Can you clarify this point?

functions/src/modules/email.functions.ts
anantakrishna: Maybe I'm overthinking. The flag might be needed in case we need to add a notification that should not be sent… As in importing from some old system, as we done for SQR, do you remember? Just in case, I want to be absolutely sure that the notification would not be sent twice.

functions/src/modules/email.functions.ts
ahmad-elbatanouni: Right, added a new line to push found templates to the object.

functions/src/modules/email.functions.ts
```diff
@@ -91,10 +166,13 @@ export const sendNotificationEmail = functions.database.ref('/email/notification
...
+        id = await getTemplateId(templateName);
+        emailTemplates[templateName] = { id };
+    }
 
-    await helpers.sendEmail(data.to, data.bcc, id, data.params);
+    if (!data['sentTimestamp'])
+        await sendEmail(data.to, data.bcc, data.replyTo, id, data.params);
 
-    return snapshot.ref.update({ sent: true });
+    return snapshot.ref.update({ sentTimestamp: Math.round((new Date()).getTime() / 1000) });
```
anantakrishna: You're updating this property even if the email was not sent this time. So that's incorrect.
And isn't there any other way of getting the timestamp? Like [this](https://firebase.google.com/docs/reference/android/com/google/firebase/Timestamp)

ahmad-elbatanouni: I found this `admin.database.ServerValue.TIMESTAMP`
It's just a placeholder value for auto-populating the current timestamp. It doesn't contain the actual timestamp. The database will replace this placeholder when it executes the command.

anantakrishna: Good.

PR #25, opened 2018-11-22
Add SE Allotment view

DadImScared:


DadImScared:
[SE: Allotment](https://trello.com/c/S2qky6S0/53-se-allotment)

.gitignore
```diff
@@ -23,3 +23,9 @@ yarn-error.log*
 *.sw*
 
 .directory
+
+/functions/lib
+
+// config for local functions
+audio-seva-team*.json
```
anantakrishna: I don't think that adding a specific project name into git config is a good idea, What kind of a file is it?

frontend/src/mixins/Lists.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, Vue } from "vue-property-decorator";
+
+@Component
+export default class Lists extends Vue {
+  lists: string[] = [];
+  listsBasePath: string = "files";
```
anantakrishna: For sound editing we're working on tasks, not files.

frontend/src/utility.ts
```diff
@@ -2,6 +2,8 @@ import _ from "lodash";
 import moment from "moment";
 import { ICount } from "@/types/DataTable";
 
+export const filteredStatus = ["Lost", "Opted out", "Incorrect", "Duplicate"];
+
```
anantakrishna: This statuses are actually belonging to the users only. So why not to move it into `UsersByRole`?

frontend/src/router.ts
```diff
@@ -43,7 +43,8 @@ export const router = new Router({
           path: "se/",
           component: () => import("@/views/SE/SE.vue"),
           children: [
-            { path: "", component: () => import("@/views/SE/Tasks.vue") }
+            { path: "", component: () => import("@/views/SE/Tasks.vue") },
+            { path: "allotment", component: () => import("@/views/SE/Allotment.vue") }
```
anantakrishna: Let's make this nested path as a verb: `allot`.

frontend/src/views/SE/Allotment.vue
```diff
@@ -0,0 +1,176 @@
...
+  };
+  selectedList: string = "";
+  submissionStatus: string = "";
+  usersRole = "SE";
+  tasks: any[] = [];
+  isLoadingLists: boolean = false;
+  isLoadingTasks: boolean = false;
+  listsBasePath: string = "/sound-editing/tasks";
+
+  soundQualityRating = {
```
anantakrishna:
```suggestion
  soundQualityRatingColor = {
```

frontend/src/mixins/UsersByRole.ts
```diff
@@ -0,0 +1,34 @@
+import { Component, Vue } from "vue-property-decorator";
+import fb from "@/firebaseApp";
+import { filteredStatus } from "@/utility";
+
+@Component
+export default class UserByRole extends Vue {
+  users: any[] = [];
+  usersBindName: string = "users";
+  usersRole: string = "";
```
anantakrishna: Maybe better to set it to `null` as a default value?

.gitignore
DadImScared: The file is from https://firebase.google.com/docs/functions/local-emulator#set_up_admin_credentials_optional

frontend/src/mixins/Lists.ts
DadImScared: The vue component that uses the mixin can override one or more of the properties of the mixin  https://vuejs.org/v2/guide/mixins.html#Option-Merging. Also I realize now it should be named something generic to the firebase shallow query since there's no reason it should be restricted to lists

frontend/src/mixins/UsersByRole.ts
DadImScared: It seems much easier to deal with the falsy value of the type rather than null. For example with the empty array this line `v-if="tasks && tasks.length"` becomes `v-if="tasks.length"`

.gitignore
anantakrishna: Do you mean that `audio-seva-team*.json` is a service account key file that you save into the project directory and thus add to `gitignore`? In this case you should save the file somewhere outside the project folder. The idea is that the source code (including `gitignore` file) should not include any artifact of the test firebase project.

frontend/src/mixins/Lists.ts
anantakrishna: So are you going to make it more generic? I could not get this.

anantakrishna: Oh I see that you've already done this.

.gitignore
anantakrishna: So please remove this.

PR #26, opened 2018-11-23
SE: Upload processing

ahmad-elbatanouni:
1. Send a notification to the coordinator
2. Update the Task

### DIFF
--- a/functions/src/modules/SE.functions.ts
+++ b/functions/src/modules/SE.functions.ts
@@ -1,0 +1,61 @@
+import * as functions from 'firebase-functions';
+import * as  admin from 'firebase-admin';
+
+const db = admin.database();
+
+/////////////////////////////////////////////////
+//
+//          UploadProcessing (Storage Trigger)
+//      1. Send a notification to the coordinator
+//      2. Update the Task
+//
+/////////////////////////////////////////////////
+
+export const uploadProcessing = functions.storage.object()
+.onFinalize(async object => {    
+
+    const filePath = object.name;
+    
+    if (object.name.slice(-4).toLowerCase() !== 'flac') {
+        console.error("Something is wrong with the file extension!")
+        return -1;
+    }
+
+    const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
+
+    if (!filePath.match(filePathRegex)) {
+        console.error("Something is wrong with the file path!")
+        return -1;
+    }
+    
+    const match = filePathRegex.exec(filePath);
+    const list = match[1], fileNameFlac = match[2];
+    const taskId = fileNameFlac.slice(0, -5); // strip the trailing ".flac" extenstion
+
+
+    // 1. Send a notification to the coordinator
+    const coordinator = functions.config().coordinator; // TO
+    const taskRef = await db.ref(`/sound-editing/tasks/${list}/${taskId}`).once('value');
+    const task = taskRef.val();
+    const replyTo = task.restoration.assignee.emailAddress;
+
+    db.ref(`/email/notifications`).push({
+        template: 'se-upload',
+        to: coordinator.emailAddress,
+        replyTo,
+        params: { task }
+    });
+
+
+    // 2. Update the Task
+    let taskRestorationUpdate = {
+        status: 'Uploaded',
+        timestampLastVersion: Math.round((new Date()).getTime() / 1000)
+    };
+
+    if (!task.restoration.timestampFirstVersion)
+        taskRestorationUpdate['timestampFirstVersion'] = Math.round((new Date()).getTime() / 1000);
+
+
+    return taskRef.ref.child('restoration').update(taskRestorationUpdate);    
+});
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -24,6 +24,9 @@ try {
 **********************************************/
 import * as SQR from './modules/SQR.functions';
 import * as ReportingSync from './modules/reporting.sync.functions';
+import * as SE from './modules/SE.functions';
 import * as Authorization from "./modules/Authorization.functions";
 
 export { SQR, ReportingSync, Authorization };
+
+export { SE }


ahmad-elbatanouni:
[SE: Upload processing](https://trello.com/c/r0CHM9Yi/28-se-upload-processing)

functions/src/modules/SE.functions.ts
```diff
@@ -0,0 +1,61 @@
...
+//      2. Update the Task
+//
+/////////////////////////////////////////////////
+
+export const uploadProcessing = functions.storage.object()
+.onFinalize(async object => {    
+
+    const filePath = object.name;
+    
+    if (object.name.slice(-4).toLowerCase() !== 'flac') {
```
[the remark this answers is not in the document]
ahmad-elbatanouni: Used the mimetype `contentType` instead

functions/src/modules/SE.functions.ts
```diff
@@ -0,0 +1,61 @@
...
+    const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
+
+    if (!filePath.match(filePathRegex)) {
+        console.error("Something is wrong with the file path!")
+        return -1;
+    }
+    
+    const match = filePathRegex.exec(filePath);
+    const list = match[1], fileNameFlac = match[2];
+    const taskId = fileNameFlac.slice(0, -5); // strip the trailing ".flac" extenstion
```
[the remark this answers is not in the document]
ahmad-elbatanouni: there's an attribute on the file object called `contentType` ... do you think it's better to modify the regex to match only files ending in **.flac**?

functions/src/modules/SE.functions.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
+
+    if (!filePath.match(filePathRegex)) {
+        console.error("Something is wrong with the file path!")
+        return -1;
+    }
+    
+    const match = filePathRegex.exec(filePath);
+    const list = match[1], fileNameFlac = match[2];
```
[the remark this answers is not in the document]
ahmad-elbatanouni: I tried using this (package)[https://www.npmjs.com/package/named-regexp-groups] but it seems to have some issues.
Do you have a particular suggestion .. or should I move on testing more packages ?

functions/src/modules/SE.functions.ts
```diff
@@ -16,15 +16,13 @@ export const uploadProcessing = functions.storage.object()
...
-
     const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
 
-    if (!filePath.match(filePathRegex)) {
-        console.error("Something is wrong with the file path!")
+    if (!filePath.match(filePathRegex)) 
+        return -1;    
+
+    if (object.contentType === 'audio/flac') {
+        console.error("Something is wrong with the file extension!")
```
anantakrishna: The message should mention the file type, not an extension. The file can have FLAC content type and another extension.

functions/src/modules/SE.functions.ts
```diff
@@ -50,11 +47,11 @@ export const uploadProcessing = functions.storage.object()
     // 2. Update the Task
     let taskRestorationUpdate = {
         status: 'Uploaded',
-        timestampLastVersion: Math.round((new Date()).getTime() / 1000)
+        timestampLastVersion: admin.database.ServerValue.TIMESTAMP
     };
 
     if (!task.restoration.timestampFirstVersion)
-        taskRestorationUpdate['timestampFirstVersion'] = Math.round((new Date()).getTime() / 1000);
+        taskRestorationUpdate['timestampFirstVersion'] = admin.database.ServerValue.TIMESTAMP;
```
anantakrishna: Can't we use a property assignment syntax here?

functions/src/modules/SE.functions.ts
```diff
@@ -34,14 +32,13 @@ export const uploadProcessing = functions.storage.object()
```
anantakrishna: Use named capture groups to get the list and the task id here.

functions/src/modules/SE.functions.ts
```diff
@@ -16,15 +16,13 @@ export const uploadProcessing = functions.storage.object()
...
-        console.error("Something is wrong with the file extension!")
-        return -1;
-    }
-
     const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
 
-    if (!filePath.match(filePathRegex)) {
-        console.error("Something is wrong with the file path!")
+    if (!filePath.match(filePathRegex)) 
+        return -1;
```
anantakrishna: Just check for the path prefix here to filter out the events. Regex should be more specific and should used after this simple check.

functions/src/modules/SE.functions.ts
```diff
@@ -16,15 +16,13 @@ export const uploadProcessing = functions.storage.object()
 
     const filePath = object.name;
     
-    if (object.name.slice(-4).toLowerCase() !== 'flac') {
-        console.error("Something is wrong with the file extension!")
-        return -1;
-    }
-
     const filePathRegex = /sound-editing\/restored\/(\w+)\/(\w+)/g;
```
anantakrishna: `\w` is not good here. Test for task ID and the file extension in this regex.

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: It's not supported yet, we need to use a package .. I tried using this [package](https://www.npmjs.com/package/named-regexp-groups) but it seems to have some issues.
Do you have a particular suggestion .. or should I move on testing more packages ?

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: It's not a property of the **taskRestorationUpdate** object, so the compiler would complain if used  as a property.
Also I'm adding it only the first time the `restoration` sub-node is updated .. so I can't add a  placeholder property of the same name to the **taskRestorationUpdate** object.

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: Ok, but the grouping still needs to be based on the whole name `list-serial.flac` .. right ?

functions/src/modules/SE.functions.ts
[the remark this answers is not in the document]
anantakrishna: Oh, I see, ECMAScript does not support named capture groups. Never mind then.

PR #27, opened 2018-12-01
Nested navigation drawer

DadImScared:


DadImScared:
[Nested navigation drawer](https://trello.com/c/WW1xqRi6/58-nested-navigation-drawer)

frontend/src/App.vue
```diff
@@ -64,12 +79,8 @@ export default {
     return {
       appTitle: "Audio Seva Backend",
       sidebar: false,
+      subNavigation: ["List", "Allot", "Statistics"],
```
anantakrishna: Is it possible to restructure the routes (make them nested) and get sub-routes from there rather than hard-coding them here?

DadImScared: It appears possible but will not update the routes if addRoutes is used to add dynamic routes.

anantakrishna: We are not going to use dynamic routes, so please refactor the code to build the nested menu items automatically from the routes. Think of the fact that not all the routes are to be present in the menu. I guess some metadata should be used for the routes.

Also research a little on whether there is a specialized solution for this just not to reinvent the wheel.

PR #31, opened 2018-12-07

ahmad-elbatanouni:
As I was working before blindly( before moving to a paid plan ), I didn't test the process on firebase..
However now after testing, the normal credentials provided by the following code

```javascript
const auth = await google.auth.getClient({
        scopes: ["https://www.googleapis.com/auth/spreadsheets"]
});
...
spreadsheet.setAuthToken(auth.credentials);
```
were insufficient and always returned an `Unauthorized access` error, so I switched to using a **service account** and mentioned the details in `README.md**

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,7 @@ The following command will run a **predeploy** script to make the conversion and
 ```sh
 $ npm run deploy
 ```
+
+# Importing data from a Google spreadsheet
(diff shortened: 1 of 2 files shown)


ahmad-elbatanouni:
[SE: Importing chunks from the spreadsheet](https://trello.com/c/VG4HHpZA/54-se-importing-chunks-from-the-spreadsheet)

PR #35, opened 2018-12-10
SE: Uploading a restored file

DadImScared:


DadImScared:
[SE: Uploading a restored file](https://trello.com/c/fqP1mh10/30-se-uploading-a-restored-file)

frontend/src/App.vue
```diff
@@ -103,6 +104,13 @@ export default {
   },
   methods: {
     ...mapActions("user", ["signOut"])
+  },
+  watch: {
+    '$route' (to) {
+      if (to.path.includes("/sound-editing/upload")) {
+        this.showHeader = false;
+      }
+    }
```
anantakrishna: I don't like this approach.
Can't we use something like [this](https://forum.vuejs.org/t/multiple-master-page-in-vue-router/20270) or [this](https://stackoverflow.com/q/37838173)?

frontend/src/utility.ts
```diff
@@ -44,3 +44,23 @@ export const formatTimestamp = (value: string, item: any) => {
   const timestamp = _.get(item, value, false);
   return timestamp ? moment(timestamp).format("D.MM.YYYY") : "";
 };
+
+const taskIdPattern = "^\\w+-\\d+-\\d+";
+
+const taskId = new RegExp(taskIdPattern);
```
anantakrishna:
```suggestion
const taskIdFormat = new RegExp(taskIdPattern);
```

frontend/src/utility.ts
```diff
@@ -44,3 +44,23 @@ export const formatTimestamp = (value: string, item: any) => {
...
+const flacFileFormat = new RegExp(`${taskIdPattern}.flac`);
+
+export const validateFlacFile = ({ name: fileName, type }: any) => {
+  const isValidFormat = fileName.match(flacFileFormat);
+  if (isValidFormat && type === "audio/flac") return isValidFormat;
+
+  throw new Error(
+      type === "audio/flac" ? "The file name does not start with task ID" : "File type must be flac"
+  );
+};
```
anantakrishna: This function looks cumbersome. Please simplify. It's better to throw an error twice than combining into one conditional statement. Just for the sake of readability.

frontend/src/views/SoundEditingUpload.vue
```diff
@@ -0,0 +1,286 @@
...
+    if (taskId) {
+      const {
+        params: { uploadCode }
+      } = this.$route;
+      this.uploadFile(
+        `sound-editing/restored/uploads/${uploadCode}/${timestamp}/${taskId}.flac`,
+        file
+      );
+    }
+  }
```
anantakrishna: Why is this method called `getTask`? Please re-consider.

frontend/src/views/SoundEditingUpload.vue
```diff
@@ -0,0 +1,286 @@
...
+      ? "Less than a minute left"
+      : moment.duration(totalUploadTime, "seconds").humanize();
+  }
+
+  uploadFile(path: string, file: File) {
+    const metadata = {
+      customMetadata: {
+        fullName: file.name
+      }
+    };
```
anantakrishna: This metadata is not needed anymore, as the file names will not contain anything except Task ID. SO even the path could be constructed here out of the file name.

frontend/src/views/SoundEditingUpload.vue
```diff
@@ -0,0 +1,286 @@
...
+                <div class="dz-details">
+                    <div class="dz-size"><span data-dz-size></span></div>
+                    <div class="dz-filename"><span data-dz-name></span></div>
+                </div>
+                <div class="dz-progress"><span class="dz-upload" data-dz-uploadprogress></span></div>
+                <div class="dz-error-message"><span data-dz-errormessage></span></div>
+                <div class="dz-success-mark"><i class="fa fa-check"></i></div>
+                <div class="dz-error-mark"><i class="fa fa-close"></i></div>
+            </div>
+        `;
```
anantakrishna: Is it necessary to declare the template as a string?

frontend/src/views/SoundEditingUpload.vue
```diff
@@ -0,0 +1,286 @@
...
+  }
+}
+</script>
+
+<style scoped>
+>>> .dropzone .fileWrapper {
```
anantakrishna: Isn't `>>>` a mistake here?

DadImScared: This is a deep selector to affect children

frontend/src/views/SoundEditingUpload.vue
DadImScared: Since it's being hidden and the uploaded files displayed below it has a custom CSS class to make sure other styles don't affect it.

frontend/src/utility.ts
anantakrishna: Still I don't like it. What I'd do is to test one condition and throw error if not met, then test another condition and throw error. And in the end I'd return `true`. Would be more readable. Would you agree?

frontend/src/views/SoundEditingUpload.vue
```diff
@@ -0,0 +1,281 @@
+<template>
+  <div>
+    <h2>Sounds Engineer Upload</h2>
```
anantakrishna: Let's name the file also `SoundEngineerUpload.vue`.

frontend/src/views/SoundEditingUpload.vue
anantakrishna: Still not clear but I'm leaving it up to you. Just wondered.

PR #36, opened 2018-12-10
Adaptation of the new requirements

ahmad-elbatanouni:


ahmad-elbatanouni:
[SE: Upload processing](https://trello.com/c/r0CHM9Yi/28-se-upload-processing)

functions/src/modules/SE.functions.ts
```diff
@@ -77,3 +79,94 @@ export const processNewAllotment = functions.database
...
+.onFinalize(async object => {    
+
+    const filePath = object.name;
+
+
+    if (!filePath.startsWith('sound-editing/restored/uploads'))
+        return -1;
+    
+    
+    const filePathRegex = /sound-editing\/restored\/uploads\/(\w+)\/(\w+)\/(\w+\/)*(([\w\d]+)-\d+-?\d+?)\.flac/;
```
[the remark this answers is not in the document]
anantakrishna: Moreover, the last counter in the task ID is mandatory, so it should be `((\w+)-\d+-\d+)`

functions/src/modules/SE.functions.ts
```diff
@@ -77,3 +79,94 @@ export const processNewAllotment = functions.database
...
+    
+    const match = filePathRegex.exec(filePath);
+    if (!match) {
+        console.log("Wrong Path -- File will be deleted.");
```
anantakrishna: Here and everywhere add a file path to the log message. Unless it is somehow specified automatically.

functions/src/modules/SE.functions.ts
```diff
@@ -77,3 +79,94 @@ export const processNewAllotment = functions.database
...
+    
+
+    
+    if (!supposedAssignee.exists()) { 
+        console.log("Wrong Upload code -- File will be deleted.");
+        bucket.file(object.name).delete();
+        return -1;
+    }
+
+    // 1. Send a notification to the coordinator
```
anantakrishna: Actually the validity check starts here, sending email goes further.

functions/src/modules/SE.functions.ts
```diff
@@ -77,3 +79,94 @@ export const processNewAllotment = functions.database
...
+    const task = taskRef.val();
+
+    if (['Revise', 'Spare'].indexOf(task.restoration.status) < 0) {
+        console.log("Incorrect task status (only [Revise OR Spare] is allowed here) -- File will be deleted.");
+        bucket.file(object.name).delete();
+        return -1;
+    }
+
+    const coordinator = functions.config().coordinator; // TO
+    const replyTo = task.restoration.assignee.emailAddress;
```
anantakrishna: Once I had already commented on this — don't use one-time variables, just put this expression directly into the push below.

storage.rules
```diff
@@ -0,0 +1,9 @@
+service firebase.storage {
+  match /b/audio-seva-team-test.appspot.com/o {
```
anantakrishna: It's not proper to commit a bucket name here. So reconsider this.

anantakrishna: And it's unclear why `mp3` is mentioned below.

README.md
```diff
@@ -65,4 +65,39 @@ The following command will run a **predeploy** script to make the conversion and
...
+Then to deploy the created storage rules, run the following:
+
+```sh
+$ firebase deploy --only storage:target-name
+```
+
+So far the created **target names** are [ se_uploads ], so to reproduce the current environment setup, run the following:
+
+```sh
+$ firebase target:apply storage se_uploads audio-seva-test-uploads
```
anantakrishna: `audio-seva-test-uploads` should not be mentioned here. No real projects/buckets etc should be revealed in the source code. Just use a placeholder here: `<sound editing uploads bucket name>`.

README.md
```diff
@@ -65,4 +65,39 @@ The following command will run a **predeploy** script to make the conversion and
...
+    "rules": "./functions/storage.sound-editing.uploads.rules"
+  }]
+}
+```
+
+Then to deploy the created storage rules, run the following:
+
+```sh
+$ firebase deploy --only storage:target-name
+```
```
anantakrishna: Above just keep the commands that a developer needs to actually issue. Don't copy the theory from the documentation, better to give a link. No need describing what to add to `firebase.json` as soon as you already added this and committed. And the target name is already defined, so just describe what is to be done on the developer computer.

functions/src/modules/SE.functions.ts
```diff
@@ -91,71 +91,68 @@ export const processNewAllotment = functions.database
...
-export const uploadProcessing = functions.storage.object()
+export const uploadProcessing = functions.storage.bucket(functions.config().sound_editing.restoration.uploads.bucket_name).object()
 .onFinalize(async object => {    
 
     const filePath = object.name;
-
-
-    if (!filePath.startsWith('sound-editing/restored/uploads'))
-        return -1;
+    const seUploadsBucket = admin.storage().bucket(functions.config().sound_editing.restoration.uploads.bucket_name);
```
anantakrishna: Do we really need to define the bucket again? Can't we get it from the `object`?

functions/src/modules/SE.functions.ts
```diff
@@ -91,71 +91,68 @@ export const processNewAllotment = functions.database
...
     const filePath = object.name;
-
-
-    if (!filePath.startsWith('sound-editing/restored/uploads'))
-        return -1;
+    const seUploadsBucket = admin.storage().bucket(functions.config().sound_editing.restoration.uploads.bucket_name);    
     
-    
-    const filePathRegex = /sound-editing\/restored\/uploads\/(\w+)\/(\w+)\/(\w+\/)*(([\w\d]+)-\d+-?\d+?)\.flac/;
+    const filePathRegex = /(\w+)\/(\w+\/)*(([\w\d]+)-\d+-\d+)\.flac/;
```
anantakrishna: Use [non-capturing group](https://javascript.info/regexp-groups#non-capturing-groups-with) to skip the folders in between. This is better than addressing the matches from the end.

functions/src/modules/SE.functions.ts
```diff
@@ -91,71 +91,68 @@ export const processNewAllotment = functions.database
...
     
     const match = filePathRegex.exec(filePath);
     if (!match) {
-        console.log("Wrong Path -- File will be deleted.");
-        bucket.file(object.name).delete();
+        console.log(`Wrong Path -- ${filePath} will be deleted.`);
+        seUploadsBucket.file(object.name).delete();
```
anantakrishna: There are too many places where you delete the file. I mean in the code. Try to restructure the function in order to delete the file once in the end of processing. It may require to extract another function.

anantakrishna: Maybe throwing an error message from the validity checks would help also to log it in one `catch`.

functions/src/modules/SE.functions.ts
```diff
@@ -91,71 +91,68 @@ export const processNewAllotment = functions.database
...
-        bucket.file(object.name).delete();
+        console.log(`Incorrect task status (only [Revise OR Spare] is allowed here) -- ${filePath} will be deleted.`);
+        seUploadsBucket.file(object.name).delete();
         return -1;
     }
 
-    const coordinator = functions.config().coordinator; // TO
-    const replyTo = task.restoration.assignee.emailAddress;
+
+    // 1. Send a notification to the coordinator
```
anantakrishna: Perform the actions in the order as they described in the card.

storage.sound-editing.uploads.rules
```diff
@@ -0,0 +1,9 @@
+service firebase.storage {
+  match /b/{bucket}/o {
+    match /{uploadCode}/{file=**} {
+      allow write: if request.resource.contentType.matches('audio/flac') &&
+          request.resource.name.matches('.*\\.flac');
```
anantakrishna: Previously you said that you cannot use regex here. Isn't it a regex? If it is then why not to test for the task ID pattern as well?

PR #40, opened 2018-12-19
Fix users listing

DadImScared:


DadImScared:
[Users listing](https://trello.com/c/hUJiBzkg/57-users-listing)

frontend/src/components/Users/InlineTextEdit.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <v-edit-dialog
+    :return-value.sync="item.notes"
+    lazy
+    large
+    @save="$listeners.save(item, editPath, item.notes)"
+    @cancel="$listeners.cancel"
```
anantakrishna: Why `listeners` is here?

DadImScared: The save and cancel functions are passed as events.

anantakrishna: I mean, why it is called “listeners” in this context? I don't understand this.

anantakrishna:
Oh, I got it now. I was thinking that it is somewhat related to the
listening please of our workflow... And I was wondering why it is mentioned
here, in the users list.

PR #41, opened 2018-12-20
Full-width layout

DadImScared:


DadImScared:
[Full-width layout](https://trello.com/c/2ARkO1eu/74-full-width-layout)

PR #47, opened 2018-12-28
Copy SQR allotments and submissions back to the spreadsheet

adrienshen:
https://trello.com/c/aGJWJMRE/76-copy-sqr-allotments-and-submissions-back-to-the-spreadsheet


adrienshen:
[Copy SQR allotments and submissions back to the spreadsheet](https://trello.com/c/aGJWJMRE/76-copy-sqr-allotments-and-submissions-back-to-the-spreadsheet)

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+  'U',
+  'V',
+  'X',
+  'Y',
+  'Z',
+];
+
+type ISheetId = string;
+
+export enum ISoundQualityReportSheet {
```
anantakrishna: I believe that any specific interfaces/enums etc related to some particular spreadsheet should be moved from this service into specific module. In this case into `SQR` module. This service should be spreadsheet-agnostic, as the `google-spreadsheet` package is.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
+const gapi = require('googleapis').google;
+
+export enum GoogleScopes {
+  SpreadSheets = 'https://www.googleapis.com/auth/spreadsheets',
+}
+
+const letters: string[] = [
```
anantakrishna: This may not work in all situations. Large spreadsheets have column numbering `AA`, `AB` etc. I think you should write a function that constructs the `A1` notation of the column by its sequential number.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+   * @param sheet The sheet to query from google sheets api
+   * @param limit How many rows you want including the header titles
+   */
+  public async getSheet(
+    spreadsheetId: ISheetId,
+    sheetName: IProjectSpreadSheetNames,
+    limit?: number
+  ): Promise<IGetSheetResponse> {
+    await this.connect();
+    // console.log("Range query: ", sheet + this._computeRange(limit));
```
anantakrishna: No commented code should be committed.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+    sheetName: IProjectSpreadSheetNames,
+    columnName: string
+  ) {
+    await this.connect();
+
+    // Retrieve first row of headers
+    const firstRow: any = await this.connection.spreadsheets.values.get({
+      spreadsheetId,
+      majorDimension: IMajorDimensions.Rows,
+      range: sheetName + '!A1:Z1',
```
anantakrishna: Z might not be the last column. I propose:
```suggestion
       range: sheetName + '!1:1',
```

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+    const { values }: IValueRange = data;
+    let targetedColumn;
+    const headers: any[] = values[0];
+    headers.forEach((elem, index) => {
+      if (elem === columnName) {
+        targetedColumn = letters[index];
+      }
+    });
+
+    // console.log("Column letter of File Name: ", targetedColumn);
```
anantakrishna: Commented code…

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+
+    const entireColumn: any = await this.connection.spreadsheets.values.get({
+      spreadsheetId,
+      majorDimension: IMajorDimensions.Rows,
+      range: sheetName + `!${targetedColumn}1:${targetedColumn}`,
```
anantakrishna:
```suggestion
       range: `${sheetName}!${targetedColumn}:${targetedColumn}`,
```
This might work even without specifying `1`, as `A:A` seems a valid range for spreadsheets.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+   * @param sheetName
+   * @param appendValues
+   */
+  public async updateAllotmentRow(
+    sheetId: string,
+    sheetName: IProjectSpreadSheetNames,
+    rowNumber: number,
+    updateValues: any
+  ): Promise<any> {
+    const targetedRange = sheetName + `!A${rowNumber}:Z${rowNumber}`;
```
anantakrishna: Again, Z is not the last column. Use `1:1` syntax.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  /**
+   * Updates a specific Allotment Sheet row
+   *
+   * @param sheetId Update row with known row number
+   * @param sheetName
+   * @param appendValues
+   */
+  public async updateAllotmentRow(
```
anantakrishna: This specific method should be abstracted as well.

functions/src/services/GoogleSheet.ts
anantakrishna: Check and remove all the commented code, I will not point out each case.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+    });
+
+    // Update this row with new merged data from database
+    // Object.keys(updateValues).forEach((elem, index) => {
+    //   targetRow.data.values[0][index] = targetRow.data.values[0][index];
+    // });
+    const resource = {
+      values: [
+        [
+          targetRow.data.values[0][0], // "Days passed"
```
anantakrishna: Is it possible to update only certain cells without updating the whole row with current values like this?

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,358 @@
...
+    let no = 0;
+    tableToSearch.forEach((elem, index) => {
+      if (elem[column] === rowValueToSearch) {
+        found = elem;
+        no = index;
+      }
+    });
+    return {
+      row: found,
+      range: `A${no + 2}:N`,
```
anantakrishna: Some specific magic is here.

functions/src/utils/parsers.ts
```diff
@@ -0,0 +1,46 @@
...
+  audioDescriptionList.forEach((elem: IAudioDescription, index: number) => {
+    multiline = multiline
+    + `${elem.beginning}-${elem.ending}:${elem.type} -- ${elem.description}`
+    + ((audioDescriptionList.length === (index + 1)) ? "" : "\n");
+  });
+  return multiline;
+}
+
+export function createUpdateLink(token: string): string {
+  return `http://purebhakti.info/audioseva/form/sound-quality-report?token=${token}`;
```
anantakrishna: This particular function is needed only in SQR module, why not to move it there?

functions/src/utils/parsers.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function spreadsheetDateFormat(date: number): string {
+  if (!date) {
+    return "-";
+  }
+  return format(date*1000, "MM/DD/YYYY");
```
anantakrishna: I'm not sure that this will work in all cases. As each spreadsheet has its own date format. Moreover, dates are stored in google sheets as a float number, please research this and make this code reliable.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +383,112 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+
+      const gsheets = new GoogleSheet();
+      const allotmentFileNames = await gsheets.getColumn(
+        spreadsheetId,
+        ISoundQualityReportSheet.Allotments,
+        'File Name'
+      );
+
+      let rowNumber;
+      allotmentFileNames.forEach((file, index) => {
```
anantakrishna: Why not `.indexOf`?

functions/src/services/GoogleSheet.ts
adrienshen: 👍 Alright.

functions/src/services/GoogleSheet.ts
adrienshen: Will remove.

functions/src/services/GoogleSheet.ts
adrienshen: Will look into this and see if there's a better way.

functions/src/services/GoogleSheet.ts
adrienshen: Understood.

functions/src/utils/parsers.ts
adrienshen: This works for all inputs that are timestamped which is the case for the sheets I was working on (Allotments and Submissions Database), and the output is "MM/DD/YYYY" string which is what I observed in the Example Spreadsheet here: https://docs.google.com/spreadsheets/d/11dlTfbkuWHlaVQFiZIFr1cdRDilqVB-shJXZWEEHcW0/edit#gid=1103170360

I tested this in Allotments and Submissions and the data string is currently working. The dates look like strings to me. Where are you getting floats from? You mean in the database? Yes, the input is float/number.

I was also thinking that all dates in the database would be timestamp format. This function is only for converting database time stamp to "MM/DD/YYYY" strings for the Spreadsheet. When on the Spreadsheet, this string will shows the calendar when clicked as expected.

adrienshen:
@anantakrishna I fixed most of what easily could be fixed. A couple remaining suggestions, awaiting your response to see if it's worth the time to modify. Eg. Not updating only certain cells and avoid updating the entire row with current values.

functions/src/services/GoogleSheet.ts
adrienshen: It doesn't seem like there's a way to specifically target cells by column name, so would likely need to calculate specific ranges for each individual cell if we want to do this without repopulating existing values.

functions/src/services/GoogleSheet.ts
anantakrishna: If I understand correctly, moving the types to a separate file is a best practice. At least we follow this approach in frontend.

functions/src/services/GoogleSheet.ts
anantakrishna: OK, no problem, no need to spend more time on this.

functions/src/modules/SQR.functions.ts
```diff
@@ -8,6 +8,20 @@ const lodash = require('lodash');
...
+  spreadsheetDateFormat,
+  withDefault,
+  commaSeparated,
+  EMPTY_VALUE,
+} from '../utils/parsers';
+
+export enum ISoundQualityReportSheet {
+  Allotments = 'Allotments',
+  Submissions = 'Submissions',
+  Statistics = 'Statistics',
```
anantakrishna: I think we don't need Statistics sheet at all.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+  comments: string;
+  name: string;
+  email_address: string;
+}
+
+/**
+ * On creation of a new allotment record id, update and sync data values to Google Spreadsheets
+ *
+ */
+export const syncAllotments = functions.database
```
anantakrishna: Let's call it
```suggestion
export const exportAllotmentToSpreadsheet = functions.database
```

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+        timestampDone,
+        followUp,
+      } = soundQualityReporting;
+      const updateResults = await gsheets.updateAllotmentRow(
+        spreadsheetId,
+        ISoundQualityReportSheet.Allotments,
+        rowNumber,
+        {
+          days_passed: EMPTY_VALUE,
+          date_given: spreadsheetDateFormat(timestampGiven),
```
anantakrishna: Check whether moment.js supports this conversion out of the box.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+          notes: withDefault(notes),
+          language: commaSeparated(languages),
```
anantakrishna: Doesn't lodash support this out of the box?

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+          status: soundQualityReporting.status,
+          file_name: fileName,
+          devotee: withDefault(assignee.name),
+          email: withDefault(assignee.emailAddress),
+          phone: EMPTY_VALUE,
+          location: EMPTY_VALUE,
+          date_done: spreadsheetDateFormat(timestampDone),
+          follow_up: withDefault(followUp),
+          list: withDefault(listName),
```
anantakrishna: We should not update list, it is immutable.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+          serial: EMPTY_VALUE,
```
anantakrishna: Serial is immutable as well.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+        spreadsheetId,
+        ISoundQualityReportSheet.Allotments,
+        rowNumber,
+        {
+          days_passed: EMPTY_VALUE,
+          date_given: spreadsheetDateFormat(timestampGiven),
+          notes: withDefault(notes),
+          language: commaSeparated(languages),
+          status: soundQualityReporting.status,
+          file_name: fileName,
```
anantakrishna: File name is not a subject of change.

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+          email: withDefault(assignee.emailAddress),
+          phone: EMPTY_VALUE,
+          location: EMPTY_VALUE,
+          date_done: spreadsheetDateFormat(timestampDone),
+          follow_up: withDefault(followUp),
+          list: withDefault(listName),
+          serial: EMPTY_VALUE,
+        }
+      );
+      console.log('Update results: ', updateResults.data);
```
anantakrishna: Can we use `console.error` if there was some error?

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+    }
+  );
+
+
+interface IAudioDescription {
```
anantakrishna:
```suggestion
interface IAudioChunkDescription {
```

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+  ending: string; // h:mm:ss
+  type: string;
+  description: string;
+}
+
+/**
+ * Used for Unwanted Parts and Sound Issues to create multi-line comments
+ *
+ */
+export function formatMultilineComment(audioDescriptionList: IAudioDescription[]) {
```
anantakrishna: Can't we add a method to the `IAudioDescription` interface (or a derived class) to format a single description, and then join them with `\n`?

functions/src/modules/SQR.functions.ts
```diff
@@ -370,3 +384,171 @@ export const importSpreadSheetData = functions.https.onRequest(
...
+    + ((audioDescriptionList.length === (index + 1)) ? "" : "\n");
+  });
+  return multiline;
+}
+
+/**
+ * On creation of a new submission record id, update and sync data values to Google Spreadsheets
+ *
+ */
+export const syncSubmissions = functions.database
```
anantakrishna:
```suggestion
export const exportSubmissionToSpreadsheet = functions.database
```

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,292 @@
+const gapi = require('googleapis').google;
+
+export enum GoogleScopes {
+  SpreadSheets = 'https://www.googleapis.com/auth/spreadsheets',
+}
+
+type ISheetId = string;
+
+// Create other SpreadSheet Interfaces and add to IProjectSpreadSheetNames
```
anantakrishna: This comment is not clear. And the interface name is also not self-documenting.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,292 @@
...
+// enum IOtherReportSheet {  }
```
anantakrishna: Looks like commented code left.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,292 @@
...
+    spreadsheetId: ISheetId,
+    sheetName: IProjectSpreadSheetNames,
+    columnName: string
+  ): Promise<any>;
+  public abstract findRowWithColumnValue(
+    column: string,
+    rowValueToSearch: string | any,
+    tableToSearch: any
+  ): Promise<any>;
+  public abstract updateAllotmentRow(
```
anantakrishna: This method cannot belong to an abstract interface, and even to the `GoogleSheets` implementation. Allotments is specific to a particular spreadsheet, our wrapper should not be aware of this. Wrapper (service) is only operating sheets, columns, rows, headers of the columns and values.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,292 @@
...
+  majorDimension: IMajorDimensions;
+  values: IRowValues[];
+}
+
+interface IGetSheetResponse {
+  headers: any[];
+  rows: any[];
+}
+
+abstract class SpreadSheet {
```
anantakrishna: I doubt that we need this level of abstraction. Really, there is no chance we will use anything besides Google Sheets as we already have legacy system running on it. Please consider.

functions/src/services/GoogleSheet.ts
```diff
@@ -0,0 +1,292 @@
...
+    updateValues: any
+  ): Promise<any>;
+}
+
+type IRowValues = string[] | number[];
+
+export default class GoogleSheets extends SpreadSheet {
+  protected connection: any;
+
+  protected async connect() {
```
anantakrishna: You are calling this method from each other method. Shouldn't we add some protection from connecting again and again?

functions/src/services/GoogleSheet.ts
anantakrishna: But the name and the function body still is not abstracted, it is tightly coupled with the allotments sheet.

functions/src/utils/parsers.ts
anantakrishna: Remove this from parsers! It's SQR-specific.

functions/src/utils/parsers.ts
```diff
@@ -0,0 +1,22 @@
+import { format } from "date-fns";
+
+export const EMPTY_VALUE = "-";
```
anantakrishna: How does it work? Why a hyphen is used? Is it a magic value?

adrienshen: Hyphen is the default. It can be a blank space instead or anything.

functions/src/modules/SQR.functions.ts
adrienshen: Removed "Statistics" sheet

functions/src/services/GoogleSheet.ts
adrienshen: Removed.

functions/src/services/GoogleSheet.ts
adrienshen: Return condition added.

functions/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "main": "lib/index.js",
   "dependencies": {
+    "date-fns": "^1.30.1",
```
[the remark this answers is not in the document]
adrienshen: Switched from datefns to moment.

functions/src/modules/SQR.functions.ts
```diff
@@ -8,6 +8,20 @@ const lodash = require('lodash');
 
 const db = admin.database();
 import * as helpers from './../helpers';
+import GoogleSheet from '../services/GoogleSheet';
```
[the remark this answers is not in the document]
adrienshen: Renamed

functions/src/modules/SQR.functions.ts
adrienshen: Renamed.

functions/src/utils/parsers.ts
anantakrishna: Do I understand correctly that there will be a hyphen in the spreadsheet if the value is empty in the database?

adrienshen:
Make the changes to make `GoogleSheets{}` as generalized as possible, removing all sheet specific logic to modules instead.

functions/src/modules/SQR.functions.ts
anantakrishna: Yes, something like that.

functions/src/modules/SQR.functions.ts
```diff
@@ -438,50 +395,40 @@ export const syncAllotments = functions.database
...
-          phone: EMPTY_VALUE,
-          location: EMPTY_VALUE,
-          date_done: spreadsheetDateFormat(timestampDone),
-          follow_up: withDefault(followUp),
-          list: withDefault(listName),
-          serial: EMPTY_VALUE,
-        }
-      );
+
+      const updateResults = await gsheets.updateRow(rowNumber, {
```
anantakrishna: I think it's better to fetch the row, update some of its properties and then pass it to `updateRow`.

functions/src/modules/SQR.functions.ts
```diff
@@ -438,50 +395,40 @@ export const syncAllotments = functions.database
...
+        "Days passed": spreadsheetDateFormat(timestampGiven),
```
anantakrishna: This column is auto calculated in the spreadsheet, we should not update it.

functions/src/modules/SQR.functions.ts
```diff
@@ -438,50 +395,40 @@ export const syncAllotments = functions.database
...
+        "Date Given": withDefault(notes),
```
anantakrishna: Mismatch. Date Given is not Notes.

functions/src/modules/SQR.functions.ts
```diff
@@ -438,50 +395,40 @@ export const syncAllotments = functions.database
...
+        "Language": commaSeparated(languages),
+        "Status": soundQualityReporting.status,
+        "File Name": fileName,
+        "Devotee": withDefault(assignee.name),
+        "Email": withDefault(assignee.emailAddress),
+        "Date Done": spreadsheetDateFormat(timestampDone),
+        "Follow Up": withDefault(followUp),
+        "List": withDefault(listName),
```
anantakrishna: Don't update list!

functions/src/services/GoogleSheets.ts
```diff
@@ -0,0 +1,199 @@
...
+    return this.connection;
+  }
+
+  /**
+   * Query for a specific google sheets within a spreadsheet
+   *
+   * @param sheet The sheet to query from google sheets api
+   * @param limit How many rows you want including the header titles
+   */
+  public async getSheet(limit?: number): Promise<any> {
```
anantakrishna: Let's call it `getRows`.

functions/src/services/GoogleSheets.ts
```diff
@@ -0,0 +1,199 @@
...
+    return afterUpdate;
+  }
+
+  /**
+   * Add a new row to specified spread sheet
+   *
+   * @param sheetId The sheet to query from google sheets api
+   * @param appendValues Data values to add to Google Sheets
+   */
+  public async appendRow<T>(appendValues: T): Promise<any> {
```
anantakrishna: What is the use of generic here, and why it was not used in `updateRow`?

functions/src/utils/parsers.ts
```diff
@@ -1,19 +1,14 @@
-import { format } from "date-fns";
-
-export const EMPTY_VALUE = "-";
-
-export function createUpdateLink(token: string): string {
-  return `http://purebhakti.info/audioseva/form/sound-quality-report?token=${token}`;
-}
+// import { format } from "date-fns";
```
anantakrishna: Guess what?

functions/src/services/GoogleSheets.ts
adrienshen: The generic is for if the end user wants to pass in a interface for strong typing, VSCode completion. For example you could: `gsheets.appendRow<IAllocationRows>(appendValues)`

We could use a generic for `updateRow` as well. Currently it's just set to `any`.

functions/src/utils/parsers.ts
adrienshen: Meant to backspace, not sure how it got commented out.

functions/src/services/GoogleSheets.ts
adrienshen: Changed.

functions/src/modules/SQR.functions.ts
adrienshen: Alright, then I need to pull the existing row.

functions/src/modules/SQR.functions.ts
adrienshen: You . mean like this?

```
      const existingRow: any = await gsheets.getRow(rowNumber);

      const updateResults = await gsheets.updateRow(rowNumber, {
        "Date Given": spreadsheetDateFormat(timestampGiven),
        "Notes": withDefault(notes),
        "Language": commaSeparated(languages),
        "Status": soundQualityReporting.status,
        "File Name": fileName,
        "Devotee": withDefault(assignee.name),
        "Email": withDefault(assignee.emailAddress),
        "Date Done": spreadsheetDateFormat(timestampDone),
        "Follow Up": withDefault(followUp),
        // Must be updated with existing value or it will be wiped by wrapper
        "List": existingRow["List"],
        "Serial": existingRow["Serial"],
      });
```

adrienshen: Or maybe like this is cleaner?
```
      const existingRow: any = await gsheets.getRow(rowNumber);
      existingRow["Date Given"] = spreadsheetDateFormat(timestampGiven);
      existingRow["Notes"] = withDefault(notes);
      existingRow["Language"] = commaSeparated(languages);
      existingRow["Status"] = soundQualityReporting.status;
      existingRow["File Name"] = fileName;
      existingRow["Devotee"] = withDefault(assignee.name);
      existingRow["Email"] = withDefault(assignee.emailAddress);
      existingRow["Date Done"] = spreadsheetDateFormat(timestampDone);
      existingRow["Follow Up"] = withDefault(followUp);
      const updateResults = await gsheets.updateRow(rowNumber, existingRow);
```

anantakrishna:
Just please resolve the conflict, then I'll merge it.

PR #48, opened 2018-12-28
Menu sections should be folded/unfolded independently

DadImScared:


DadImScared:
[Menu sections should be folded/unfolded independently](https://trello.com/c/4HN27NeZ/90-menu-sections-should-be-folded-unfolded-independently)

PR #49, opened 2018-12-29

ahmad-elbatanouni:


ahmad-elbatanouni:

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+                    }
+
+
+                    ///////////////////////////////
+                    // Sorting by Beginning time
+                    ///////////////////////////////
+                    let chunks = lastFile.chunks;
+
+                    // `start` attribute is added to be used in sorting
+                    chunks.forEach(element => element.start = moment.duration(element.beginning).asSeconds());
```
anantakrishna: Why can't you sort by `chunk.beginning`? I think it's identical to your artificial `start` attribute.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+                    chunks = lodash.orderBy(chunks, ['start'], ['asc']);
+
+                    // getting rid of the added `start` attribute
+                    chunks.forEach(element => delete element.start);
+
+
+
+                    ////////////////////////////////////////////////////////
+                    // Filling `continuationFrom` only for the FIRST column
```
anantakrishna: Why do you use “column” here? Cannot understand it.

anantakrishna: Moreover, why are you doing this? Is there a requirement in the task for this?

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+                    ////////////////////////////////////////////////////////
+                    for (let z = 0; z < chunks.length; z++) 
+                        if (chunks[z].continuationFrom && z !== 0)
+                            delete chunks[z].continuationFrom;
```
anantakrishna: I'm sure this can be done using `forEach`.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+                //          --> last audio file chunks were saved to firebase 
+                //          --> Reset `lastFile` object to hold the last read row 
+                //                         OR
+                // 2. last chunk is `skipped` and therefor all of the previous chunks
+                //      of the same audioFile ( the ones in the `lastFile` object ) is discarded
+                //          --> Reset `lastFile` object to hold the last row
+
+
+                lastFile.file_name = row[columnsIndex['AudioFileName']];
+                lastFile.chunks = [{
```
anantakrishna: Here you're repeating the same code as above (line 204). This is an anti-pattern. Rewrite the code so that you don't copy-paste it.

functions/src/modules/SE.functions.ts
anantakrishna: I guess why. You misunderstood the “Continuation From column should be filled only for the first chunk of the track (file)” validity check requirement. You should **verify** that the “Continuation From” column is filled only for the first chunk, whereas you just **remove** this value for all the chunks except the first one. This is wrong.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+            addedChunks: 0, 
+            emptyValues: 0, 
+            misplaced: 0, 
+            noResolution: 0, // rows with No `Resolution` or No `FidelityCheckResolution` values
+            overlappingChunks: [], // Chunks with overlapping times
+            modified: [] // Rows already found in the DB
+        };
+
+    
+    for (let i = 0; i < sheetTitles.length; i++) { // Loop throughout the sheets of interest
```
anantakrishna: Can't `forEach` be used here?

functions/src/helpers.ts
```diff
@@ -2,3 +2,103 @@
...
+      });
+
+      if (!result.data.values || result.data.values.length === 0) {
+          console.log(`No data found in sheet: ${title}`);
+          continue;
+      }
+  
+      // Getting only the sheets that have the required attributes
+      const row = result.data.values[0] 
+      if (row.indexOf('Beginning') > -1 && row.indexOf('Ending') > -1) {
```
anantakrishna: This is specific to your particular function, it's not proper to put this in helpers. Helpers should provide generic code that can be used for any other spreadsheet with another columns.

functions/src/modules/SE.functions.ts
anantakrishna: And processing a sheet should be extracted into a separate function. I had pointed out earlier that you need to re-structure the code, but you seem to have ignored that.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+        //          B. Non-existence of `Resolution` values in a row
+        //          C. Empty values in either `Beginning` or `Ending` attributes
+        //
+        //
+        ////////////////////////////////////////
+        for (const row of rows) {
+            let skip = false;
+
+
+            // [CHECK #1] Misplaced audio file entry
```
anantakrishna: Use phrase from the requirements, that would simplify testing and understanding. Try not to invent your terms and not to rephrase.

anantakrishna: And first of all you should test for the first Filter condition, which is related to the Resolutions. Because “Other rows should be skipped silently”.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,230 @@ export const processNewAllotment = functions.database
...
+                continue;
+            }
+
+
+            resolution = row[columnsIndex['Resolution']];
+            fidCheckRes = row[columnsIndex['FidelityCheckResolution']];
+
+            if (resolution && fidCheckRes) {
+                // [CHECK #2] Non-existence of `Resolution` values in a row
+                if (resolutions.indexOf(resolution) < 0 || resolutions.indexOf(fidCheckRes) < 0 || resolution != fidCheckRes) {
```
anantakrishna: 1. Don't mix two validity checks in one place, as there should be different messages in the summary for them.
1. As far as I know, `indexOf` is case-sensitive, whereas the requirement is that the resolutions should be compared case-insensitively.
1. Here you just skip the row, whereas the validity check failure should produce a warning in the log and summary email.

functions/src/modules/SE.functions.ts
anantakrishna: I though so. The question is actually just to bring your attention to this issue.

functions/src/modules/SE.functions.ts
anantakrishna: And please add `continuationTo` property as described in the data structure. It is not existent in the spreadsheet, you'll have to add it based on the `continuationFrom`.

functions/src/modules/SE.functions.ts
anantakrishna: No-no, perhaps I was not clear that time, or there was another reason for not using it. In general, `forEach` is much more readable.

functions/src/modules/SE.functions.ts
anantakrishna: You extracted the helpers. Here I'm speaking of separating the main function logic itself.

PR #52, opened 2019-01-01
Add Listening page

ojhasubham:


frontend/src/router.ts
```diff
@@ -37,6 +37,15 @@ export const router = new Router({
       path: "/sound-editing/:taskId/quality-check/feedback",
       component: () => import("@/views/QCSubmissionForm.vue")
     },
+    {
+      path: "/listen/:fileName",
+      component: () => import("@/views/Listen/ListenLayout.vue"),
```
anantakrishna: Why not `AnonymousLayout`? I was expecting you to use it.

frontend/src/router.ts
```diff
@@ -37,6 +37,15 @@ export const router = new Router({
...
+      children: [
+        {
+          path: "", component: () => import("@/views/Listen/Listen.vue")
```
anantakrishna: Can we put this view in the root of the `/views` folder? Now it looks like an artificial folder.

frontend/src/views/Listen/Listen.vue
```diff
@@ -0,0 +1,74 @@
...
+
+<style>
+</style>
+
+<script>
+export default {
+  name: "ListenAudio",
+  data: () => ({
+    fileName: null,
+    storageBaseDomainName: "test.storage.audioseva.purebhakti.info",
```
anantakrishna: No. This should not be hard-coded. Read again https://trello.com/c/59mJXDBs/73-storage-structure.

frontend/src/views/Listen/Listen.vue
```diff
@@ -0,0 +1,74 @@
...
+    baseUrl: "http://original.",
+    audioUrl: ""
+  }),
+  mounted: function() {
+    this.getAudioUrl();
+  },
+  methods: {
+    getAudioUrl() {
```
anantakrishna: Why not making it a computed property?

frontend/src/views/Listen/Listen.vue
```diff
@@ -0,0 +1,74 @@
...
+<style>
+</style>
+
+<script>
+export default {
+  name: "ListenAudio",
+  data: () => ({
+    fileName: null,
+    storageBaseDomainName: "test.storage.audioseva.purebhakti.info",
+    baseUrl: "http://original.",
```
anantakrishna: The name of this property is not correct. You can name it `storageSubdomain` and save there only `original`. But this is not a `baseUrl`. Moreover, you're using it only once, no need making it a property here at all.

frontend/src/router.ts
ojhasubham: I used AnonymousLayout but for the particular listening page, I changed it. But I will use AnonymousLayout.

frontend/src/router.ts
ojhasubham: Sure, I will put it to the root of /view

frontend/src/views/Listen/Listen.vue
ojhasubham: ok, I will get it from *process.env.VUE_APP_STORAGE_ROOT_DOMAIN*

frontend/src/views/Listen/Listen.vue
ojhasubham: I have made it in computed property

frontend/src/views/Listen/Listen.vue
ojhasubham: I have removed this property

frontend/src/views/Listen.vue
```diff
@@ -0,0 +1,68 @@
...
+</style>
+
+<script>
+export default {
+  name: "ListenAudio",
+  data: () => ({
+    fileName: null,
+  }),
+  computed: {
+    getAudioUrl: function() {
```
anantakrishna: Computed property should not start with `get`. It's just `audioUrl`.

frontend/src/views/Listen.vue
```diff
@@ -0,0 +1,68 @@
+/*
+ * sri sri guru gauranga jayatah
+ */
+<template>
+  <div>
+    <v-toolbar :clipped-left="$vuetify.breakpoint.lgAndUp" fixed app>
+      <v-toolbar-title>
+        <router-link to="/" tag="span" style="cursor: pointer">
+          <h3>Audio File Name: {{ fileName ? fileName.split('.')[0]: null }}</h3>
```
anantakrishna: It's not good to put this logic into the template. Rather create another computed property for the base name of the file.

frontend/src/views/Listen.vue
```diff
@@ -0,0 +1,68 @@
...
+
+<script>
+export default {
+  name: "ListenAudio",
+  data: () => ({
+    fileName: null,
+  }),
+  computed: {
+    getAudioUrl: function() {
+      this.fileName = this.$route.params.fileName;
```
anantakrishna: I don't think it's a good idea to change the data property inside computed property.

anantakrishna: I expect you to propose the most proper way to do it. I'm thinking that the filename can be done a separate computed property.

ojhasubham: yes I have Done with separate computed property

PR #56, opened 2019-01-02
Add search functionality to the SE listing

ojhasubham:


frontend/src/views/SE/Tasks.vue
```diff
@@ -55,9 +64,10 @@ import { formatTimestamp, getDayDifference } from "@/utility";
   components: { DataTable, SoundIssuesList }
 })
 export default class Tasks extends Vue {
-  tasks: ITasks[] = [];
+  tasks: any[] = [];
```
anantakrishna: Why removing the strong typing here?

ojhasubham: Yes, I was going to ask you about it but by mistake I forgot it to ask you about that I was getting some problem so I think to change any to make same with the user list page.

My problem was resolve without this change I will change it and push it again 

Don't worry

ojhasubham: Done

anantakrishna:
Please rebase your commit on top of the last `master`. There was some refactoring, and linting was added. After that lint your code (`npm run lint`) and format the file (Shift+Alt+F in VSCode).

PR #59, opened 2019-01-04
Fix CR Allotment form warnings

ojhasubham:


anantakrishna:
Just a side-note for the future: no need to force-push every time. You can add new commits to the same branch describing in the commit message what was done/fixed/improved. In this case, fixing other sections could be done with additional commits.

Force-push  is useful when old commits are not compliant with our standards: wrong message, not relevant accidental changes are included, etc.

PR #62, opened 2019-01-05

ahmad-elbatanouni:


ahmad-elbatanouni:

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+const validateSpreadsheetRow = (row, summary, sheetTitle) => {
+    let resolutions = ['ok', 'drop', 'duplicate', 'on hold', 'reallot', 'repeat', 'derivative'];
+    let resolution, fidCheckRes; // variables to hold values of Res and fidRes when looping through the rows
+
+    let _import = false;
+ 
+
+    // [CHECK #1] File name belongs to the list identified by the sheet name       
+    if (helpers.extractListFromFilename(row['Audio File Name']) != sheetTitle) {
+        console.log("File name belongs to a different list.");
```
anantakrishna: This should be warning, not just log. In all the cases below.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+            console.log(`Audio file name: ${row['Audio File Name']} -- Error: Resolution, or Fidelity check resolution values isn't valid`);
+            summary.invalidResoultion.push({
+                audioFileName: row['Audio File Name'],
+                resolution,
+                fidCheckRes,
+                errorMessage: `Resolution, or Fidelity check resolution values isn't valid.`
+            });
+        }
+
+        // [CHECK #3] Resolution and Fidelity Check Resolution should be equal, case-insensitively.
```
anantakrishna: We discussed, this check is not needed. Remove.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+            });
+        }
+
+        // [CHECK #6] Beginning and Ending should be filled for rows with Resolution equal to `OK`.
+        if (resolution === 'ok') {
+            _import = true;
+        }
+
+    } else // [Filter criterion #1] Rows not having (Resolution, or Fidelity check resolutin) are skipped SILENTLY
+        summary.noResolution.push({ audioFileName: row['Audio File Name'], resolution, fidCheckRes });
```
anantakrishna: No need reporting this case, as there will be hundreds and thousands of such files.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                audioFileName: row['Audio File Name'],
+                resolution,
+                fidCheckRes,
+                errorMessage: `Mismatch in Resolution and Fidelity check resolution values.`
+            });
+        }
+
+        // [CHECK #6] Beginning and Ending should be filled for rows with Resolution equal to `OK`.
+        if (resolution === 'ok') {
+            _import = true;
```
anantakrishna: Where is the check for Beginning and Ending here?

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+        scopes: ["https://www.googleapis.com/auth/spreadsheets.readonly"]
+    });
+
+    const sheets = google.sheets({ version: "v4", auth });
+    const spreadsheetId = functions.config().reporting.content.processing.spreadsheet_id;
+    let sheetscMetadata  = await helpers.buildSheetIndex(sheets, spreadsheetId);
+
+
+    let currentFile, // used to increment the chunks of an audiofile coming from multiple rows into a single object
+        summary = {
```
anantakrishna: I think that it might be more logical to group the summary around the rows themselves. Something like
```
summary = {
  validityFailures: [{
    row: {}, //data
    issues: ["misplaced", "invalidResolution", …]
  }]
}

File: functions/src/modules/SE.functions.ts [hunk 6]
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+            noResolution: [], // rows with No `Resolution` or No `FidelityCheckResolution` values
+            invalidResoultion: [],
+            mismatchedResolutions: [],
+            overlappingChunks: [], // Chunks with overlapping times
+            modified: [] // Rows already found in the DB
+        };
+
+    
+    sheetscMetadata.forEach(async sheet => {
+        if (sheet.firstRow.indexOf('Beginning') < 0 || sheet.firstRow.indexOf('Ending') < 0)
```
anantakrishna: Document this check.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                _import
+            };
+
+
+            
+            // Same file name --> add newly read chunk (row) to the `currentFile` object
+            if (currentFile.file_name === AudioFileName) 
+                currentFile.chunks.push(chunkToStore);
+            
+            // NEW FILE --> Save old file AND add the newly read chunk to `currentFile` object
```
anantakrishna: Some black magic is happening below.
Rather create another function that “flushes” the current file into the database.
The purpose of the `chunks` variable is unclear.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                }
+
+                currentFile.file_name = AudioFileName;
+                currentFile.chunks = [chunkToStore];
+            }                    
+        }
+        
+
+        ////////////////////////////////
+        // Processing Sheet Audio Files
```
anantakrishna: The comment is not clear.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+        //          A. Misplaced files
+        //          B. Non-existence of `Resolution` values in a row
+        //          C. Empty values in either `Beginning` or `Ending` attributes
+        //
+        ////////////////////////////////////////
+        for (const row of rows) {
+
+            let _import = validateSpreadsheetRow(row, summary, sheet.title);            
+            
+            const AudioFileName = row['Audio File Name'];
```
anantakrishna: Variable should not begin with the capital letter.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+            ////////////////////////////////////////////
+            // [CHECK #5] Chunks of a single audio file 
+            // should not overlap each other in timing.
+            ////////////////////////////////////////////
+            let lastEndingTime = null;
+            currentFilechunks.forEach((chunk, i) => {
+
+
+                // `Beginning` value should be less than `Ending`.
+                if (chunk.beginning > chunk.ending) {
```
anantakrishna: Why not to perform this check in the `validateSpreadsheetRow ` function?

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                    console.log(`Audio file name: ${fileName} -- Error: "Beginning" value isn't less than "Ending".`);
+                    summary.overlappingChunks.push({
```
anantakrishna: This is not “Chunks of a single audio file should not overlap each other in timing” case. This is “Beginning value should be less than Ending” case.

Overlapping means that two chunks are overlapping each other. Example:

- 00:00-15:00
- 10:00-20:00

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                    console.log(`Audio file name: ${fileName} -- Error: Overlapping chunks.`);
+                    summary.overlappingChunks.push({
+                        audioFileName: fileName,
+                        errorMessage: `Overlapping chunks.`
+                    });
+                }
+
+
+                // setting the value of `lastEndingTime` to be used in the next iteration
+                if (chunk.ending)
```
anantakrishna: All the chunks should have ending time, it must be validated earlier, isn't it? So no need to check here.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+                else
+                    lastEndingTime = null;
+
+
+                //////////////////////////////////////////////////
+                // [CHECK #4] `continuationFrom` column should
+                // be filled only for the first chunk of the track.
+                //////////////////////////////////////////////////
+                if (chunk.continuationFrom && i !== 0)
+                    delete chunk.continuationFrom;
```
anantakrishna: You should not modify the data! “Should” means that if it if filled for any other chunk than for the first one, you should issue a warning in the summary email.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+
+                
+                // importing only chunks with `_import` flag set to true
```
anantakrishna: This logic is unclear.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +93,276 @@ export const processNewAllotment = functions.database
...
+            currentFilechunks = await lodash.orderBy(currentFilechunks, ['beginning'], ['asc']);
+            // console.log(currentFilechunks);
+
+            let chunksToImport = [];
+            ////////////////////////////////////////////
+            // [CHECK #5] Chunks of a single audio file 
+            // should not overlap each other in timing.
+            ////////////////////////////////////////////
+            let lastEndingTime = null;
+            currentFilechunks.forEach((chunk, i) => {
```
anantakrishna: Validating the whole file with all its chunks should be extracted to a separate function.

As I see now, we have per-row validation rules and per-file. The latter ones are spanning all the chunks. These are: overlapping chunks and `continuationFrom`. So have two functions: `validateSpreadsheetRow` and `validateFileChunks`.

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: The **chunks object** is used to group all the chunks (rows) of a single file ... 
The task card says the rows of an audio file are not guaranteed to be successive .. so I thought of using an object with the following structure

```javascript
chunks = {
    "BR-01A": [ chunk#1, chunk#2, .... ],
    "BR-03A": [ chunk#1, chunk#2, .... ],
}
```
and **after all the rows of a sheet are read**, the rows are flushed to the database.
That's why I can't flush the chunks of the **current file** whenever a row is read.

Do you suggest to flush the current file to the db whenever a new file name is encountered and later *append* any chunks that belong to a file **existing** in the db ?
@anantakrishna

ahmad-elbatanouni: Bear in mind, the chunks have to be sorted by `Beginning` time, so pushing them to the db and rereading them for sorting would be a waste of time.

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: sorry, forgot to add them to a different category

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: It's validated for **importing** .. not for **processing** ... or did I get this one wrong ?

functions/src/modules/SE.functions.ts
ahmad-elbatanouni: Here, I only import chunks that passed all the validity checks.

functions/src/modules/SE.functions.ts
anantakrishna: > The task card says the rows of an audio file are not guaranteed to be successive

I should have described this in a more clear way, sorry. The card says:

- Rows related to a particular file name are supposed to be clubbed together.
- Rows within a file name group may be out of order.
- Rows are not guaranteed to be sorted by the file name.

This means that the chunks belonging to a particular file are grouped together, but these groups may be not sorted by the file name, and the chunks may be not sorted within a group. For example:

```
BR-03A chunk 1
BR-03A chunk 2
BR-01A chunk 3
BR-01A chunk 1
BR-01A chunk 2
```

So, we can “flush the current file to the db whenever a new file name is encountered”. And if later the same file name occurs, then an error should be reported and no overwriting should occur.

functions/src/modules/SE.functions.ts
anantakrishna: There is only one “validation”.
So the pipeline is as follows:

1. Is the row finalized (have both resolutions)? Yes — proceed with it. No — skip silently.
1. Is the row valid? All the per-row checks should be applied here. Valid — proceed. No — issue a warning.
1. Are all rows of the same file finalized and valid? Yes — import all the rows of this file with “OK” resolution. No — don't import any row of this file.

Now should be more clear.

functions/src/modules/SE.functions.ts
anantakrishna: Perhaps this may be eliminated according to my previous comments.

PR #63, opened 2019-01-05
Update firebase rule

ojhasubham:


ojhasubham:
[Fix CR Allotment form warnings](https://trello.com/c/Oh0S5kjQ/113-fix-cr-allotment-form-warnings)

PR #65, opened 2019-01-07
Process registration webform data

hamzaavvan:
Fixed errors: Promises must be handled appropriately from GoogleSheets.ts and SQR.functions.ts

### DIFF
--- a/functions/src/modules/Users.functions.ts
+++ b/functions/src/modules/Users.functions.ts
@@ -10,7 +10,7 @@ const db = admin.database();
  * 
  * @function processRegistration()
  */
-export const processRegistration = functions.database
+export const restructureRegistrationData = functions.database
 .ref('/webforms/registration/{registration_id}')
 .onCreate(async (snapshot, context) => {
   const webform = snapshot.val();
--- a/functions/src/modules/Users.functions.ts
+++ b/functions/src/modules/Users.functions.ts
@@ -1,0 +1,62 @@
+import * as functions from 'firebase-functions';
+import * as admin from 'firebase-admin';
+
+const db = admin.database();
+
+/**
+ * Process registration of webform data
+ * 1. Registers user by transferring data into /users/ path
+ * 2. Check if users already exists
+ * 
+ * @function processRegistration()
+ */
+export const processRegistration = functions.database
+.ref('/webforms/registration/{registration_id}')
+.onCreate(async (snapshot, context) => {
+  const webform = snapshot.val();
+
+  const oldUser = await db.ref("/users/").orderByChild("/emailAddress")
+                  .equalTo(webform.email_address).once("value");
+
+  // If the user already exists stop the execution and log the error
+  if (oldUser.val() !== undefined) {
+    console.warn(`Warning! User already exist with email: ${webform.email_address}`)
+    return false;
+  }
+
+  const serviceRoles = {
+    "Content Reporting": "CR",
+    "Track Editing":     "TE",
+    "Sound Engineering": "SE",
+    "Quality Checking":  "QE",
+    "Fidelity Checking": "FC",
+  };
+
+  // if user email doesn't exists prepare the newUser Object
+  const newUser: any = {
+    timestamp: webform.completed,
+    emailAddress: webform.email_address,
+    phoneNumber: webform.contact_number,
+    isAvailableOnWhatsApp: webform.i_am_available_on_whatsapp === 1,
+    languages: webform.languages.reduce((result, language) => {
+      result[language] = true;
+      return result;
+    }, {}),
+    location: webform.country,
+    name: webform.your_name,
+    services: (<any>Object).values(webform.seva).join(", "),
+    experience: webform.experience,
+    influencer: webform.where_did_u_hear_about_this_seva,
+    recommendedBy: webform.recommended_by,
+    roles: webform.seva.reduce((result, service) => {
+      const code = serviceRoles[service];
+      result[code] = true;
+
+      return result;
+    }, {})
+  }
+
+  db.ref(`/users/`).push(newUser);
+
+  return true;
+});
--- a/database.rules.json
+++ b/database.rules.json
@@ -88,6 +88,9 @@
         "notes": {
           ".write": "auth.token.Coordinator === true"
         },
+        "emailAddress": {
+          ".write": "!data.exists()"
+        },
         "languages": {
           "$langeName": {
             ".validate": "newData.isBoolean()"


hamzaavvan:
[Process registration webform data](https://trello.com/c/ptt0cWCw/109-process-registration-webform-data)

functions/src/services/GoogleSheets.ts
```diff
@@ -103,8 +103,8 @@ export default class GoogleSheets {
   }
 
   public async getRow(rowNumber: number): Promise<any> {
-    this.connect();
-    this.getHeaders();
+    await this.connect();
+    await this.getHeaders();
```
anantakrishna: This may be a good suggestion, however this is not related to your current task. See point 5 of https://trello.com/c/rkJKKl5d/107-git-conventions.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
 
-/////////////////////////////////////////////////
-//          OnNewAllotment (DB create and update Trigger)
-//      1. Mark the files in the database --> { status: "Given" }
-//              Function --> updateFilesOnNewAllotment
-//
-//      2. Send an email to the assignee to notify them of the new allotments
-//              Function --> sendEmailOnNewAllotment
-/////////////////////////////////////////////////
-export const updateFilesOnNewAllotment = functions.database
```
anantakrishna: Why are you removing this?

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
-        timestampDone: null,
-      });
+/**
+ * Process registration of webform data
+ * 1. Registers user by transferring data into /users/ path
+ * 2. Check if users already exists
+ * 
+ * @function processRegistration()
+ */
+export const processRegistration = functions.database
```
anantakrishna: This function is not belonging to the SQR module.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+.ref('/webforms/registration/{registration_id}')
+.onCreate(async (snapshot, context) => {
+  const user = snapshot.val();
 
-      if (sqrError == undefined) {
-        // if Successful FILE Update, update the ALLOTMENT accordingly
+  const olduser = await db.ref("/users/").orderByChild("/emailAddress")
```
anantakrishna: Do we need an index on this field? If it doesn't exist then it should be added to the `database.rules.json`.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
-        else
-          db.ref(`/sqr/allotments/${newDocKey}`).update({
-            filesAlloted: true,
-            sendNotificationEmail: true,
-          });
-      }
-    })
+  // If the user already exists stop the execution and log the error
+  if (olduser.val() !== undefined) {
+    console.log(`Error! User already exist with email: ${user.email_address}`)
```
anantakrishna: This should be a warning, and not just a log message. And this is not an error, this is a warning.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+    return false;
   }
-)
+
+  // if user email doesn't exists prepare the new_user Object
+  const new_user: any = {
+    timestamp: moment.format('x'),
+    emailAddress: user.email_address,
+    phoneNumber: user.contact_number,
+    isAvailableOnWhatsApp: user.i_am_available_on_whatsapp,
```
anantakrishna: Will “Yes” automatically transform into a boolean value?

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+    location: user.country,
+    name: user.your_name,
+    services: (<any>Object).values(user.seva).join(", "),
```
anantakrishna: I cannot understand the logic of this transformation. I think that there is no transformation needed at all.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+    experience: user.experience,
+    influencer: user.where_did_u_hear_about_this_seva,
+    recommendedBy: user.recommended_by,
+    status: user.status,
```
anantakrishna: There is no `status` in the source data.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+  }
+
+  /**
+   * Assigning user roles to new_user 
+   * Get the first letters of service name and assign it to @var new_user's roles
+   */
+  user.services.map(name => name.split(' ').map(piece => piece[0]).join(''))
```
anantakrishna: I believe this can be done inline in the object construction above… However if it is not easy to do, no need wasting time.

The only thing is that there is no `user.services` property.

functions/src/modules/SQR.functions.ts
```diff
@@ -20,52 +20,58 @@ export enum ISoundQualityReportSheet {
...
+ * Process registration of webform data
+ * 1. Registers user by transferring data into /users/ path
+ * 2. Check if users already exists
+ * 
+ * @function processRegistration()
+ */
+export const processRegistration = functions.database
+.ref('/webforms/registration/{registration_id}')
+.onCreate(async (snapshot, context) => {
+  const user = snapshot.val();
```
anantakrishna: To avoid confusion I would suggest
```suggestion
  const webform = snapshot.val();
```

functions/src/modules/SQR.functions.ts
```diff
@@ -520,7 +526,8 @@ export const exportSubmissionsToSpreadsheet = functions.database
         token,
         unwantedParts,
       } = snapshot.val();
-      gsheets.appendRow({
+      
+      await gsheets.appendRow({
```
anantakrishna: Irrelevant change.

functions/src/modules/SQR.functions.ts
hamzaavvan: You're right its not `user.services`. And it can't be done up in the Object that's why i did that separately.

functions/src/services/GoogleSheets.ts
hamzaavvan: I was getting error during linting that's why i fixed that.

functions/src/modules/SQR.functions.ts
hamzaavvan: @anantakrishna I did that so we can log the warning message.

anantakrishna: Your comment is irrelevant to my comment. Am I missing something?

functions/src/modules/SQR.functions.ts
anantakrishna: Anyhow, this transformation is not needed here.

functions/src/modules/SQR.functions.ts
hamzaavvan: I can't follow this: 
> Do we need an index on this field? 

Can you please elaborate?

functions/src/modules/SQR.functions.ts
hamzaavvan: @anantakrishna See this is the error we get if we remove that type transformation.

image

anantakrishna: OK, I mistakenly was thinking that `webform.seva` is a string. No issues here. Thank you.

functions/src/modules/SQR.functions.ts
anantakrishna: I've checked that we have an index on `emailAddress` property of user in `database.rules.json`, so no issues here. That's what I meant.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Assigning user roles to new_user 
+   * Get the first letters of service name and assign it to @var new_user's roles
+   */
+  webform.seva.map(name => name.split(' ').map(piece => piece[0]).join(''))
+              .map(role => new_user.roles[role] = true);
+  
+  // assigning user languages to new_user Object 
+  webform.languages.map(lang => new_user.languages[lang] = true);
```
anantakrishna: I believe you should use [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) instead of `map` here. Map is not intended to have any side effect, and it is supposed to return a mapped array, which we don't need. Whereas you can use `reduce` to accumulate the roles in the object, and in this case you will be able to assign it directly to `newUser`.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+  
+  db.ref(`/users/`).push(new_user);
+
+  return 1;
```
anantakrishna: Just curious, why 1? Why not true?

hamzaavvan: I don't think it makes such impact, but i'll change it to `true` because its a good practice. Thanks!

image

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // If the user already exists stop the execution and log the error
+  if (oldUser.val() !== undefined) {
+    console.warn(`Warning! User already exist with email: ${webform.email_address}`)
+    return false;
+  }
+
+  // if user email doesn't exists prepare the newUser Object
+  const newUser: any = {
+    timestamp: moment.format('x'),
```
anantakrishna: Use `completed` from the webform data.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+    emailAddress: webform.email_address,
+    phoneNumber: webform.contact_number,
+    isAvailableOnWhatsApp: webform.i_am_available_on_whatsapp === 1,
+    languages: webform.languages.reduce((map, lang) => {
```
anantakrishna:
```suggestion
    languages: webform.languages.reduce((result, language) => {
```

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+      map[lang] = true;
+      return map;
+    }, {}),
+    location: webform.country,
+    name: webform.your_name,
+    services: (<any>Object).values(webform.seva).join(", "),
+    experience: webform.experience,
+    influencer: webform.where_did_u_hear_about_this_seva,
+    recommendedBy: webform.recommended_by,
+    roles: webform.seva.reduce((map, role) => {
```
anantakrishna:
```suggestion
    roles: webform.seva.reduce((result, service) => {
```

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,55 @@
...
+      // Get the first letters of service name
+      const code = role.split(' ').map(name => name[0].join(''));
```
anantakrishna: I had already pointed it out that you cannot take first letters to get a role name because the service name can be changed, and not in all cases it will work. There should be a dictionary. Like this:

Content Reporting → CR
Something Else → XX

hamzaavvan: So you mean that in some cases role name would not be the first letters of words like this one `Sound Quality Reporting = SQR`

BTW this is the demonstration i.e result of the script:
image

anantakrishna: Yes, I mean exactly this.

functions/src/modules/Users.functions.ts
```diff
@@ -25,27 +24,40 @@ export const processRegistration = functions.database
     return false;
   }
 
+  const _services = {
```
anantakrishna: Using an underscore is not a standard (AFAIK). Please rename to `serviceRoles`.
And why not to inverse the dictionary? It would simplify logic below. It would become just `result[serviceRoles[service]] = true;`. Isn't it? The only thing is to check for existence in a similar way as you do.

PR #66, opened 2019-01-08
Sound Editing tasks compilation

DadImScared:


DadImScared:
[Sound Editing tasks compilation](https://trello.com/c/1Yj4InOY/79-sound-editing-tasks-compilation)

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
     });
     return 1;
   });
+
+const serialRegex = "^[a-zA-Z]+-\\d+";
```
anantakrishna: I'm sure we already have a regex for the file names to extract list, serial number and the other parts. Please look around and use that regex. If it doesn't exist then create it, but outside SE module, as it is common to all the modules.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+
+const updateChunk = async (path: string, taskId: string) => {
+  return await db.ref(path).update({
+    taskId
+  });
+};
+
+const makeTaskId = (fileName: string, index: number) => {
```
anantakrishna: I doubt it's worth extracting this function that is used only once in the code below.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+    continuationTo,
+    continuationFrom,
+    processingResolution = "",
+    taskId,
+    beginning,
+    ending
+  } = snapshot.val();
+  if (continuationTo || processingResolution.toLowerCase() !== "ok" || taskId) return;
+  const chunkDuration = ending - beginning;
+  const chunkTaskId = makeTaskId(fileName, index);
```
anantakrishna: Logic here is not correct.
Index of the chunk within a file cannot be used directly as a counter for the taskId.
See for example `ML1-069-3` task in the picture below. This task originates from the first chunk of the `ML1-069B` file, but it has counter `3`.
Naming scheme

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+  let duration;
+
+  if (continuationFrom) {
+    const response = await db
+        .ref(`/sound-editing/chunks/${listId}/${continuationFrom}`)
+        .orderByKey()
+        .limitToLast(1).once("value");
+    const chunks = response.val();
```
anantakrishna: To make it clear
```suggestion
    const previousFileChunks = response.val();
```

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+    const lastChunk = chunks[chunks.length - 1];
```
anantakrishna:
```suggestion
    const previousChunk = chunks[chunks.length - 1];
```

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+    if (lastChunk.continuationTo === fileName) {
```
anantakrishna: This is very important check. Better to negate it and issue a warning message, aborting further processing of the current file.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
     });
     return 1;
   });
+
+const serialRegex = "^[a-zA-Z]+-\\d+";
+
+const updateChunk = async (path: string, taskId: string) => {
```
anantakrishna: I don't feel that extracting this function helps understanding the code where it is called. I suggest doing the update inline below.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+).onCreate(async (snapshot, { params: { fileName, listId, index } }) => {
+  const {
+    continuationTo,
+    continuationFrom,
+    processingResolution = "",
+    taskId,
+    beginning,
+    ending
+  } = snapshot.val();
+  if (continuationTo || processingResolution.toLowerCase() !== "ok" || taskId) return;
```
anantakrishna: Here we cannot blindly ignore chunks with `continuationTo`. What if the “continuation” chunk was added first, and this chunk was added after that? We should check whether continuation chunk already exists in the database, and if so — we should create the task from these two chunks.

anantakrishna: However check that there is no task created from that “continuation” chunk till the date.

functions/src/modules/SE.functions.ts
```diff
@@ -88,3 +88,58 @@ export const processNewAllotment = functions.database
...
+          `/sound-editing/chunks/${listId}/${continuationFrom}/${chunks.length - 1}`,
+          chunkTaskId
+      );
+      await updateChunk(`/sound-editing/chunks/${listId}/${fileName}/${index}`, chunkTaskId)
+    }
+  }
+  else {
+    duration = chunkDuration;
+  }
+  await db.ref(`/sound-editing/tasks/${listId}/${chunkTaskId}`).update({
```
anantakrishna: Isn't there a `create` method? As it reads illogical that we update the node whereas we are supposed to add/create it.

functions/src/modules/SE.functions.ts
anantakrishna: And it will be more clear to name it
```suggestion
  const taskId = makeTaskId(fileName, index);
```

anantakrishna: It is important to test that this task id does not exist in the database, otherwise we should issue an error message and exit.

PR #68, opened 2019-01-10
SQR Submission Email Template

DadImScared:


DadImScared:
[SQR: Submission Email Template](https://trello.com/c/fD9lTxuF/70-sqr-submission-email-template)

email/sqr-submission.html
```diff
@@ -1,6 +1,55 @@
...
+        {{ allotment.daysPassed }},
+        {{ allotment.language }}
+        {% if allotment.fileName == params.submission.fileName %}
+        <strong> -- CURRENT SUBMISSION</strong>
+        {% endif %}
+      </li>
+      {% endfor %}
+    </ul>
+  </div>
+  <a href="https://audio-seva.firebaseapp.com/sqr/allot?emailaddress={{ params.submission.author.emailAddress }}">Allot the next set</a><br>
```
anantakrishna: The domain name of the frontend should be passed as a parameter. This will allow us to use the same template for development and production environments.

email/sqr-submission.html
```diff
@@ -1,6 +1,55 @@
...
+  <p>
+    Please update your previous submission using this update link: http://purebhakti.info/audioseva/form/sound-quality-report?token={{params.submission.token}}.<br>Please note &#8212; do not make a new submission using this link. For a new submission, use the link in the allotment email.<br>
+  </p>
   <p><b>A. Audio File Name:</b> {{ params.submission.fileName }}</p>
   <p><b>B. Unwanted parts to be cut:</b><br></p>
```
anantakrishna: I guess there is no need in `br` here.

anantakrishna: And in Sound Issues as well.

functions/src/modules/SQR.functions.ts
```diff
@@ -278,6 +278,21 @@ export const processSubmission = functions.database
...
+        .equalTo(submission.author.emailAddress).once("value")
+    ).val();
+
+    currentSet = Object.entries(currentSet).map(
+        ([fileName, { languages, soundQualityReporting: { timestampGiven, status } }]: any) => ({
+          fileName,
+          status,
+          timestampGiven: timestampGiven ? moment(timestampGiven).format("M/D/YYYY") : "",
+          daysPassed: timestampGiven ? moment().diff(timestampGiven, "days") : "Not available",
+          language: languages.join(", ")
```
anantakrishna: I believe this can be done in the template, it has `join` filter.

functions/src/modules/SQR.functions.ts
```diff
@@ -278,6 +278,21 @@ export const processSubmission = functions.database
   if (allSubmissionsSnapshot.exists()) {
     const allSubmissions = allSubmissionsSnapshot.val();
 
+    let currentSet = (await db
+        .ref(`/files/${list}`)
+        .orderByChild("soundQualityReporting/assignee/emailAddress")
+        .equalTo(submission.author.emailAddress).once("value")
+    ).val();
+
+    currentSet = Object.entries(currentSet).map(
```
anantakrishna: I'm not completely understand this construct. Can't we use `forEach` to iterate the objects and alter them as needed instead of mapping into another array?

functions/src/modules/SQR.functions.ts
```diff
@@ -286,7 +301,7 @@ export const processSubmission = functions.database
       params: {
         submission,
         fileData,
-        currentSet: [],
+        currentSet: currentSet,
```
anantakrishna: Can be shortened, isn't it?

email/sqr-submission.html
```diff
@@ -0,0 +1,76 @@
...
+
+    {% if params.submission.cancellation.audioProblem %}
+      <p>Unable to play or download the audio {{ params.submission.comments }}</p>
+    {% endif %}
+
+    {% if params.submission.changed != params.submission.completed %}
+      <p>This is an updated submission!</p>
+    {% endif %}
+
+    {% if params.fileData.soundQualityReporting.assignee.emailAddress != params.submission.author.emailAddress %}
```
anantakrishna: This condition is related only to the first warning in the following block. Other warnings should have its own condition. So no need in `div` here, `p` is enough.

functions/src/modules/SQR.functions.ts
DadImScared: All of them are altered because they need fileName as well as other properties.

email/sqr-submission.html
```diff
@@ -1,6 +1,55 @@
...
+        </p>
+
+        <p>Status of file is {{ params.fileData.soundQualityReporting.status }}</p>
+
+        <p>Audio file name {{ params.submission.fileName }} is not found in the Backend!</p>
+      </div>
+    {% endif %}
+
+    {% if params.fileData.soundQualityReporting.status == "Given" %}
+      <p>It's time to allot!</p>
```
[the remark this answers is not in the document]
anantakrishna: You know, now I'm thinking that it might be better to produce all the warnings in the function and pass the warnings as `params.warnings`. This way it will be more convenient to add new warnings later without changing the template.

So please move all this logic into the function.

email/sqr-submission.html
anantakrishna: This is not fixed.

functions/src/modules/SQR.functions.ts
```diff
@@ -209,6 +209,40 @@ export const restructureExternalSubmission = functions.database
...
+  if (!response) warnings.push(`Audio file name ${fileName} is not found in the backend!`);
+  let shouldShowStatusWarning = true;
+  if (status === "Given") {
+    for (const allotment of currentSet) {
+      if (allotment.status === "Given") {
+        shouldShowStatusWarning = false;
+        break;
+      }
+    }
+    if (shouldShowStatusWarning) warnings.push("It's time to allot!");
```
anantakrishna: The trigger for this warning is if the current set has only one file in the “Given” status. I'm sure it's possible to simplify the  algorithm.

functions/src/modules/SQR.functions.ts
```diff
@@ -278,16 +312,39 @@ export const processSubmission = functions.database
...
+        .ref(`/files/${list}`)
+        .orderByChild("soundQualityReporting/assignee/emailAddress")
+        .equalTo(submission.author.emailAddress).once("value")
+    ).val();
+
+    const currentSet = [];
+
+    Object.entries(allotments).forEach(
+        ([fileName, { languages, soundQualityReporting: { timestampGiven, status } }]: any) => {
+          currentSet.push({
```
anantakrishna: The idea behind using `forEach` was not to create new array, but to alter items in the same array. Effectively you just need to add `daysPassed` property to the allotment and (probably) modify `timestampGiven` property.

functions/src/modules/SQR.functions.ts
```diff
@@ -278,16 +312,39 @@ export const processSubmission = functions.database
...
+            fileName,
+            status,
+            timestampGiven: timestampGiven ? moment(timestampGiven).format("M/D/YYYY") : "",
```
anantakrishna: Are you sure this is not possible to format in the template itself? Sorry if I was asking this already.

DadImScared: I was unable to format with the template.

PR #69, opened 2019-01-10

ahmad-elbatanouni:


ahmad-elbatanouni:

ahmad-elbatanouni:
This is not a final version,
I added the following comment on Trello 
> Maximum number of rows per request (1000 rows) wasn't accounted for in the google-spreadsheet wrapper ...
It indeed has a limit parameter .. but it has no start parameter..
I'm setting the limit for now to 999 until I get a confirmation whether I can edit the wrapper or not.
Also there's a problem when the number of columns is greater than 26 (i.e needs more than one character for its name)

anantakrishna:
Replied there. Also please see my valuable comments in #62. It was a good idea to name the branch properly and start a new pull request.

PR #73, opened 2019-01-14
Import users from the spreadsheet

adrienshen:
Tested on this sheet: https://docs.google.com/spreadsheets/d/1dpGOD0IWFcwzXLjXNFXehsnqYtok5aqOWJHbLwXboIs/edit#gid=10719879

`sheetId` stored in `user.spreadsheet_id`

Updates `https://console.firebase.google.com/u/1/project/audio-seva-team-test/database/audio-seva-team-test/data/users` if the email does not already exist.


adrienshen:
[Import users from the spreadsheet](https://trello.com/c/GMOmG5PT/56-import-users-from-the-spreadsheet)

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,89 @@
+import * as functions from 'firebase-functions';
+import * as admin from 'firebase-admin';
+import GoogleSheet from '../services/GoogleSheets';
+import { DataSnapshot } from 'firebase-functions/lib/providers/database';
```
anantakrishna: I'm not sure this looks fine… Do we really need to import deeply from inside the library? What is the benefit and is it a best practice?

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      return {
+        notes: withDefault(elem[RegistrationColumns.Details]),
+        status: elem[RegistrationColumns.Status],
+        timestamp:
+          new Date(elem[RegistrationColumns.Timestamp]).getTime() / 1000,
+        name: elem[RegistrationColumns.Name],
+        location: elem[RegistrationColumns.Country],
+        emailAddress: elem[RegistrationColumns.EmailAddress],
+        phoneNumber: elem[RegistrationColumns.PhoneNumber],
+        whatsApp: elem[RegistrationColumns.WhatsApp],
```
anantakrishna: Wrong property name.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,89 @@
...
+        languages: elem[RegistrationColumns.Languages].split(', '),
```
anantakrishna: Languages are also boolean-valued objects, not arrays.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,89 @@
...
+        services: elem[RegistrationColumns.Services],
+        roles: {
+          CR: elem[Roles.CR] === Decision.Yes,
```
anantakrishna: I believe this can be done in a more elegant way. If nothing comes in 5-10 minutes, then keep it as is.

functions/src/modules/Users.functions.ts
```diff
@@ -0,0 +1,89 @@
...
+          SQR: elem[Roles.SQR] === Decision.Yes,
+        },
+        experience: elem[RegistrationColumns.Experience],
+        influencer: elem[RegistrationColumns.Influencer],
+        recommendedBy: elem[RegistrationColumns.RecommendedBy],
+      };
+    });
+
+    await readyForDatabaseUpdate.forEach(async (spreadsheetRecord: any) => {
+      const Users = db.ref('/users');
```
anantakrishna: Local variables should be camel-cased.

anantakrishna:
The PR title should be equal to the card title, not card slug.

functions/src/modules/Users.functions.ts
adrienshen: Okay, will fix.

functions/src/modules/Users.functions.ts
adrienshen: Ok, already changed.

functions/src/modules/Users.functions.ts
adrienshen: Even references to database table names? I only made it capital to tell the difference between normal variables and Database tables. Changing to camelCase now.

functions/src/modules/Users.functions.ts
adrienshen: I believe we can make it more elegant by using a reduce, although it seems easier to understand this way.

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+
+/**
+ * 1. Scan the Registrations Google Sheet and exports any new users to database
+ * 2. Do not import if user email already exists
+ *
+ */
+export const importUserRegistrationData = functions.https.onRequest(
+  async (req: functions.Request, res: functions.Response) => {
+    const gsheets: GoogleSheet = new GoogleSheet(
+      functions.config().user.spreadsheet_id,
```
anantakrishna:
```suggestion
      functions.config().registrations.spreadsheet_id,
```
And please document this in `README`.

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+        status: elem[RegistrationColumns.Status],
+        timestamp:
+          new Date(elem[RegistrationColumns.Timestamp]).getTime() / 1000,
+        name: elem[RegistrationColumns.Name],
+        location: elem[RegistrationColumns.Country],
+        emailAddress: elem[RegistrationColumns.EmailAddress],
+        phoneNumber: elem[RegistrationColumns.PhoneNumber],
+        isAvailableOnWhatsApp: elem[RegistrationColumns.WhatsApp],
+        languages: elem[RegistrationColumns.Languages]
+          .split(', ')
```
anantakrishna: I'd split just by comma, and then trim the tokens. This will help us in case there is no space after comma.

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+          .reduce((acc: any, language: string) => {
```
anantakrishna:
```suggestion
          .reduce((result: any, language: string) => {
```

functions/src/modules/Users.functions.ts
adrienshen: Done.

functions/src/modules/Users.functions.ts
adrienshen: Good idea. Done.

functions/src/index.ts
```diff
@@ -28,5 +28,6 @@ import * as ReportingSync from './modules/reporting.sync.functions';
 import * as Email from './modules/email.functions';
 import * as SE from './modules/SE.functions';
 import * as Authorization from './modules/Authorization.functions';
+import * as User from './modules/Users.functions';
```
anantakrishna: 1. Name the file `User.functions.ts`.
2. Move all the functions from `Authorization` file into `User` as they are also related to user. `Authorization` should be removed, of course.

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+export const importUserRegistrationData = functions.https.onRequest(
+  async (req: functions.Request, res: functions.Response) => {
+    const gsheets: GoogleSheet = new GoogleSheet(
+      functions.config().registrations.spreadsheet_id,
+      'Registrations'
+    );
+
+    const registrationRows = await gsheets.getRows();
+
+    const readyForDatabaseUpdate = registrationRows.map((elem: any) => {
```
anantakrishna:
```suggestion
    const readyForDatabaseUpdate = registrationRows.map((row: any) => {
```

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+      return {
+        notes: withDefault(elem[RegistrationColumns.Details]),
+        status: elem[RegistrationColumns.Status],
+        timestamp:
+          new Date(elem[RegistrationColumns.Timestamp]).getTime() / 1000,
```
anantakrishna: Is it possible to use `momentjs` here to make the transformation more self-documenting? Currently it's not obvious what's happening.

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+        recommendedBy: elem[RegistrationColumns.RecommendedBy],
+      };
+    });
+
+    await readyForDatabaseUpdate.forEach(async (spreadsheetRecord: any) => {
+      const usersRef = db.ref('/users');
+      await usersRef
+        .orderByChild('emailAddress')
+        .equalTo(spreadsheetRecord['emailAddress'])
+        .once('value', async (snapshot: functions.database.DataSnapshot) => {
```
anantakrishna: It seems that we are using callback here. Can't it be replaced by `await` to make the logic more straightforward?

functions/src/modules/Users.functions.ts
```diff
@@ -1,62 +1,162 @@
...
+  }
+);
+
 /**
  * Process registration of webform data
  * 1. Registers user by transferring data into /users/ path
  * 2. Check if users already exists
- * 
+ *
  * @function processRegistration()
```
anantakrishna: As soon as you are formatting the whole file, please make this documentation fit the actual function name.

functions/src/modules/Users.functions.ts
adrienshen: Yes, we can use moment.unix()

functions/src/modules/Users.functions.ts
adrienshen:
```
      const snapshot = await usersRef
        .orderByChild('emailAddress')
        .equalTo(spreadsheetRecord['emailAddress'])
        .once('value');
      if (snapshot.exists() === false) {
        // No such email exist, create new record
        await usersRef.push(spreadsheetRecord);
      }
```

functions/src/modules/Users.functions.ts
adrienshen: Seems now the name has changed to restructureRegistrationData() ?

I'm confused, but I change the comments to reflect this.

functions/src/modules/User.functions.ts
```diff
@@ -0,0 +1,214 @@
...
+  async (req: functions.Request, res: functions.Response) => {
+    const gsheets: GoogleSheet = new GoogleSheet(
+      functions.config().registrations.spreadsheet_id,
+      'Registrations'
+    );
+
+    const registrationRows = await gsheets.getRows();
+
+    // Removes the header rows
+    registrationRows.splice(0, 1);
```
anantakrishna: Isn't it that `getRows()` return only data rows without headers? It should be the case. Our sheets service should take care of the headers.

@ahmad-elbatanouni please comment here as well.

ahmad-elbatanouni: Yes, my bad, I mistakenly removed the line that removes the headers 
```javascript
this.headers = values.shift();
```
because it was removing the first row of each 1000 rows of each request.

anantakrishna:
Ahmad, can you fix this issue in this branch and adjust the code in all the
places that call this method? Adrien is not available until Monday, and
this is the last remark I have. So if you can do it, I could merge this
pull request today.

ahmad-elbatanouni:
@anantakrishna , edited and tested .. you can merge now

PR #75, opened 2019-01-16
Absence of SendInBlue key prevents from deploying functions

hamzaavvan:
Configure email functions to deploy in the absence of SendInBlue api key


hamzaavvan:
[Absence of SendInBlue key prevents from deploying functions](https://trello.com/c/9ras6d7c/93-absence-of-sendinblue-key-prevents-from-deploying-functions)

hamzaavvan:
[Absence of SendInBlue key prevents from deploying functions](https://trello.com/c/9ras6d7c/93-absence-of-sendinblue-key-prevents-from-deploying-functions)

functions/src/modules/email.functions.ts
```diff
@@ -3,7 +3,15 @@ import * as admin from 'firebase-admin';
 
 // SendInBlue Helper Imports
 import * as SibApiV3Sdk from 'sib-api-v3-sdk';
-const sendInBlueSecretKey = functions.config().send_in_blue.key;
+
+// Abort if send_in_blue api key is not set
+if (!functions.config().send_in_blue) {
+  console.error(`Error! Send in blue api key is not set.`);
+  process.abort();
```
anantakrishna: I'm afraid that this is not proper to do it in the global context. Will it issue an error even if the function from other module is called? I'm supposing that this global code is executed on every function invocation.

hamzaavvan: I was also thinking the same, and finding a way to sort it out. Let me search, maybe I would be able to inject the code before every request, just like a middleware.

hamzaavvan: OK i found out that we can use middleware to check every function's request. But it requires express to be installed in order to use middleware.

Ref: https://firebase.google.com/docs/functions/http-events

anantakrishna: I don't think that Express Middleware would help us, as the functions are not HTTP-triggered. Rather I'd suggest that we have some function that initializes and returns the API instance. And that function can be called from whatever function needs to use it.

hamzaavvan: What if I create a new method `checkIfApiKeyExists()` within `SibApiV3Sdk` object and call it before every `apInstance` within the functions.

PR #77, opened 2019-01-18
87 allow editing some fields in sqr listing

DadImScared:


DadImScared:
[Allow editing some fields in SQR listing](https://trello.com/c/7r0POVxR/87-allow-editing-some-fields-in-sqr-listing)

PR #78, opened 2019-01-19
Rewrite importing SQR spreadsheet

ahmad-elbatanouni:
Changes to the `GoogleSheets` service created by **Adrien** still needs refinement ... This is just an initial draft.

### DIFF
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -305,8 +305,8 @@ const parseAudioChunkRemark = (string) => {
    */
   const regex = /((Entire file)|(.*?)–(.*)):(.*)—(.*)/g;
   const tokens = [];
-  let matches;
-  while (matches = regex.exec(string)) 
+  let matches = regex.exec(string);
+  while (matches) {
     tokens.push({
       entireFile: matches[2]? true : null,
       beginning: matches[2]? null : matches[3],
@@ -314,6 +314,8 @@ const parseAudioChunkRemark = (string) => {
       type: matches[5].trim(),
       description: matches[6].trim()
     });
+    matches = regex.exec(string);
+  }
   
   return tokens;
 }
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -297,7 +297,6 @@ export const processSubmission = functions.database
  * Parse "Sound Issue" or "Unwanted parts" string to extract its different parts
  * 
  * @param string "Sound Issue" or "Unwanted parts" string to parse
- * @param array array of "Sound Issues" or "Unwanted parts"
  */
 const parseAudioChunkRemark = (string) => {
   /**
@@ -305,7 +304,7 @@ const parseAudioChunkRemark = (string) => {
    * 'g' flag is used to match one or more occurences of the pattern
    */
   const regex = /((Entire file)|(.*?)–(.*)):(.*)—(.*)/g;
-  let tokens = [];
+  const tokens = [];
   let matches;
   while (matches = regex.exec(string)) 
     tokens.push({
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -394,8 +394,8 @@ export const importSpreadSheetData = functions.https.onRequest(
       const list = helpers.extractListFromFilename(fileName);
       const allotment = {
         status: row['Status'] || null,
-        timestampGiven: row['Date Given'] ? (new Date(row['Date Given'])).getTime() : null,
-        timestampDone: row['Date Done'] ? (new Date(row['Date Done'])).getTime() : null,
+        timestampGiven: row['Date Given'] ? (new Date(row['Date Given'])).getTime() / 1000 : null,
+        timestampDone: row['Date Done'] ? (new Date(row['Date Done'])).getTime() / 1000 : null,
         assignee: {
           emailAddress: row['Email'] || null,
           name: row['Devotee'] || null
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -308,7 +308,7 @@ export const importSpreadSheetData = functions.https.onRequest(
     //     Submissions
     ////////////////////////
     const submissionsSheet = new GoogleSheets(spreadsheetId, ISoundQualityReportSheet.Submissions);
-    const submissionsRows = await submissionsSheet.__getRows();
+    const submissionsRows = await submissionsSheet.getRows();
     for (const row of submissionsRows) {
       const audioFileName = row['Audio File Name'];
       const match = audioFileRegex.exec(audioFileName);
@@ -370,7 +370,7 @@ export const importSpreadSheetData = functions.https.onRequest(
     ////////////////////////
 
     const allotmentsSheet = new GoogleSheets(spreadsheetId, ISoundQualityReportSheet.Allotments);
-    const allotmentsRows = await allotmentsSheet.__getRows();
+    const allotmentsRows = await allotmentsSheet.getRows();
 
     for (const row of allotmentsRows) {
       if (!row) continue;
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -339,9 +339,8 @@ export const importSpreadSheetData = functions.https.onRequest(
         type: unwantedpartsMatch[3],
         description: unwantedpartsMatch[4]
       };
+
       const submission = {
-        fileName: row['Audio File Name'],
-        changed: row['Changed'] || null,
         completed: row['Completed'] || null,
         created: row['Created'] || null,
         comments: row['Comments'],
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -346,10 +346,10 @@ export const importSpreadSheetData = functions.https.onRequest(
 
       const submission = {
         completed: row['Completed'] || null,
-        created: row['Created'] || null,
+        created: row['Completed'] || null,
         comments: row['Comments'],
         soundIssues,
-        soundqualityrating: row['Sound Quality Rating'],
+        soundQualityRating: row['Sound Quality Rating'],
         unwantedParts,
         duration: {
           beginning: row['Beginning'],
--- a/functions/src/modules/SQR.functions.ts
+++ b/functions/src/modules/SQR.functions.ts
@@ -290,6 +290,47 @@ export const processSubmission = functions.database
     return 1;
   });
 
+
+
+
+/**
+ * Parse "Sound Issue" or "Unwanted parts" string to extract its different parts
+ * 
+ * @param string "Sound Issue" or "Unwanted parts" string to parse
+ * @param array array of "Sound Issues" or "Unwanted parts"
+ */
+const parseSI_UP = (string, array) => {
+  /**
+   * Regex to parse the value "Sound Issues" & "Unwanted Parts"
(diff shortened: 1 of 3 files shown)


ahmad-elbatanouni:
[Rewrite importing SQR spreadsheet](https://trello.com/c/mhfg8Edy/94-rewrite-importing-sqr-spreadsheet)

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+    const spreadsheetId = functions.config().sqr.spreadsheet_id;
+
+    // Regex for validating file names in both sheets
+    const audioFileRegex = /(\w+)-\d+.*/;
+
+    ////////////////////////
+    //     Submissions
+    ////////////////////////
+    const submissionsSheet = new GoogleSheets(spreadsheetId, ISoundQualityReportSheet.Submissions);
+    const submissionsRows = await submissionsSheet.__getRows();
```
anantakrishna: I'm confused by the method name. Looks like it is not intended to be called from outside. @adrienshen can you comment here?

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+    for (const row of submissionsRows) {
+      const audioFileName = row['Audio File Name'];
+      const match = audioFileRegex.exec(audioFileName);
+      if (!match) continue;
+
+      const list = match[1];
+      // const serial = row['Submission Serial'];
```
anantakrishna: Don't keep commented code.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+
+      const fileRef = await db.ref(`/files/${list}/${audioFileName}/soundQualityReporting`).once('value');
+      const audioFile = fileRef.val();
+      let token;
+      if (fileRef.exists()) {
+        token = audioFile.token;
+      } else {
+        console.warn(`Token not found!`)
```
anantakrishna: This check and message do not correlate. All this block above is not clear, please rewrite it and make comments on WHY are we doing this.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+        continue;
+      }
 
       const regex = /(.*?)–(.*):(.*)—(.*)/g;
```
anantakrishna: This variable should be renamed to reflect the purpose of this regex.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-      const soundissuesMatch = regex.exec(row['soundissues']);
-      const unwantedpartsMatch = regex.exec(row['unwantedparts']);
+      const soundissuesMatch = regex.exec(row['Sound Issues']);
```
anantakrishna: Keep in mind, sound issues now can have “Entire File” entry. Added such an example into the sample sheet (row 6).

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+      const unwantedpartsMatch = regex.exec(row['Unwanted Parts']);
```
anantakrishna: Variables should be camel-cased. And above also.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-          description: soundissuesMatch[4],
-        },
-        unwantedparts = {
-          beginning: unwantedpartsMatch[1],
-          ending: unwantedpartsMatch[2],
-          type: unwantedpartsMatch[3],
-          description: unwantedpartsMatch[4],
-        };
+      if (!soundissuesMatch || !unwantedpartsMatch) {
+        console.warn(`"Sound Issues" or "Unwanted Parts" are not in the correct format`);
```
anantakrishna: For what file? It will be impossible to deal with this warning.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-          name: row['name'],
-          emailAddress: row['emailAddress'],
-        },
-        fileName: row['audiofilename'],
-        changed: row['changed'],
-        completed: row['completed'],
-        created: row['created'],
-        comments: row['comments'],
+        fileName: row['Audio File Name'],
+        changed: row['Changed'] || null,
```
anantakrishna: We don't have such a column in the spreadsheet.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+        completed: row['Completed'] || null,
+        created: row['Created'] || null,
```
anantakrishna: We don't have such a column in the spreadsheet.

functions/src/modules/SQR.functions.ts
anantakrishna: Token should be extracted from the Update Link column and overwritten in the database.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-        author: {
-          name: row['name'],
-          emailAddress: row['emailAddress'],
-        },
-        fileName: row['audiofilename'],
-        changed: row['changed'],
-        completed: row['completed'],
-        created: row['created'],
-        comments: row['comments'],
+        fileName: row['Audio File Name'],
```
anantakrishna: File name should not be saved in the submission anymore as it is specified in the path.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-        }
+    for (const row of allotmentsRows) {
+      if (!row) continue;
+
+      const fileName = row['File Name'];
+      const match = audioFileRegex.exec(fileName);
+
+      if (!match) continue;
+
+      const fileNameHasForbiddenChars = fileName.match(/[\.\[\]$#]/g);
```
anantakrishna: Are all forbidden characters specified here according to the Firebase documentation?

anantakrishna: And again, don't use one-time variables.

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+      if (fileNameHasForbiddenChars) {
+        console.warn(`File name has forbidden characters that can't be used as a node name.`)
```
anantakrishna: Specify the file name in the message, otherwise what is the use of the message?

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-            sendNotificationEmail: false, // Don't send emails if the document is read from the spreadsheet
-          };
-
-          db.ref(`/sqr/allotments`).push(allotment);
-        }
+    for (const row of allotmentsRows) {
+      if (!row) continue;
+
+      const fileName = row['File Name'];
+      const match = audioFileRegex.exec(fileName);
```
anantakrishna: Don't we have a helper function that extracts list from file name?

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
-    res
-      .status(200)
-      .send(
-        `Function was called successfully, check the Logs on Firebase to find out if something went wrong`
-      );
-  }
-);
+      const list = match[1];
+      const allotment = {
+        status: row['Status'] || null,
```
anantakrishna: How does this `||` work here and what is the purpose?

functions/src/modules/SQR.functions.ts
```diff
@@ -303,118 +299,107 @@ export const processSubmission = functions.database
...
+        timestampGiven: row['Date Given'] ? (new Date(row['Date Given'])).getTime() : null,
```
anantakrishna: Are you sure this is the correct way to convert sheets time into the unix timestamp?

functions/src/services/GoogleSheets.ts
```diff
@@ -65,6 +66,65 @@ export default class GoogleSheets {
...
+
+    return this._convertRows(values);
+  }
+
+  /**
+   * Getting the rows of the sheet in increments of 1000s
+   * as this is the MAX num of rows allowed in one call
+   * 
+   */
+  public async __getRows(): Promise<any> {
```
anantakrishna: Why a public method has such prefix in name?

functions/src/services/GoogleSheets.ts
```diff
@@ -65,6 +66,65 @@ export default class GoogleSheets {
...
   }
 
+  /**
+   * Get a number of rows (start - limit) from a specific google sheet within a spreadsheet
+   * with a maximum number of rows (limit) of 1000
+   * 
+   * @param start The row to start reading from
+   * @param limit How many rows you want
+   */
+  public async getRowsByStart(start: number, limit?: number): Promise<any> {
```
anantakrishna: I don't expect that this function should be public.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: I was intending on removing the original `getRows` and set the name of this as `getRows` .. `__getRows` is just a temporary name

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: This is to set the value of `status` to null so that it isn't written to the db when the cell value is an empty string

anantakrishna: Got it.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: I got them from a detailed error .. I'll look into the docs too to be absolutely sure.

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: it is used for parsing the values of two columns ... Is it okay to use the names of the two columns in the variable name ?

functions/src/services/GoogleSheets.ts
ahmad-elbatanouni: Ok, I replaced now the previous `getRows` with the body of the `__getRows` in this commit.
Also, made `getRowsByStart` **protected** and changed its name to  `_getRowsByStart`

ahmad-elbatanouni: @anantakrishna

functions/src/modules/SQR.functions.ts
anantakrishna: Indeed in this case they are optional, so just don't gill them.

ahmad-elbatanouni: @anantakrishna

anantakrishna: Can you see my response above?

functions/src/modules/SQR.functions.ts
ahmad-elbatanouni: @anantakrishna 
I should divide it by 1000 .. is that what you mean or something else ?

ahmad-elbatanouni: I tested it both locally and on firebase and it's working ... thought there's something wrong with the logic itself.
image

PR #81, opened 2019-02-03
Copy new CR submissions into the processing spreadsheet

ahmad-elbatanouni:
There's something that hasn't been considered .. Normally I check if `Submission serial` exists on the **target** sheet .. if so I do **not** copy the row from the **source** sheet (taking into consideration that it should be copied if `Fused Lectures` is set to **No**).

Now, what if 2 submission in the **source** sheet had the same `Submission serial` .. should I take that too into consideration?

This is **not** done automatically since importing is performed in batches, not row by row.

### DIFF
--- a/functions/src/modules/CR.functions.ts
+++ b/functions/src/modules/CR.functions.ts
@@ -1,0 +1,65 @@
+import * as functions from 'firebase-functions';
+import * as helpers from '../helpers';
+import GoogleSheets from '../services/GoogleSheets';
+import * as _ from 'lodash';
+
+export const copySubmissionsToProcessing = functions.pubsub
+  .topic('daily-tick')
+  .onPublish(async () => {
+    const onlineSheet = await new GoogleSheets(
+      functions.config().cr.submissions.spreadsheet.id
+    ).useSheet('Online');
+
+    const destSpreadsheet = new GoogleSheets(
+      functions.config().cr.processing.spreadsheet.id
+    );
+    const destinations = new Map();
+    const sheetNames = await destSpreadsheet.getSheetNames();
+
+    for (const row of await onlineSheet.getRows()) {
+      const list = helpers.extractListFromFilename(row['Audio File Name']);
+
+      if (!destinations.has(list)) {
+        /// We are not creating new sheets
+        if (sheetNames.indexOf(list) < 0) continue;
+
+        const sheet = await destSpreadsheet.useSheet(list);
+        destinations.set(list, {
+          sheet,
+          submissionSerials: await sheet.getColumn('Submission Serial'),
+          audioFileNames: await sheet.getColumn('Audio File Name'),
+          insertionsCounter: [],
+          rows: [],
+        });
+      }
+
+      const destination = destinations.get(list);
+
+      /// Skipping rows that were already copied
+      if (destination.submissionSerials.indexOf(row['Submission Serial']) > -1)
+        continue;
+
+      const originalIndex = _.sortedIndex(
+        destination.submissionSerials,
+        row['Submission Serial']
+      );
+
+      /// Correcting index according to the pending insertions
+      const correctedIndex = destination.insertionsCounter.reduce(
+        (sum: number, value: number, index: number) =>
+          sum + (index <= originalIndex ? value : 0),
+        originalIndex
+      );
+
+      destination.rows[correctedIndex] = row;
+
+      ++destination.insertionsCounter[originalIndex] ||
+        (destination.insertionsCounter[originalIndex] = 1);
+    }
+
+    /// Actually inserting the rows for each sheet
+    for (const destination of destinations.values())
+      await destination.sheet.addRows(destination.rows);
+
+    return true;
+  });
--- a/functions/src/services/GoogleSheets.ts
+++ b/functions/src/services/GoogleSheets.ts
@@ -1,29 +1,21 @@
 const gapi = require('googleapis').google;
+import Sheet from './Sheet';
 
 export enum GoogleScopes {
   SpreadSheets = 'https://www.googleapis.com/auth/spreadsheets',
 }
 
-enum IMajorDimensions {
-  Rows = 'ROWS',
-  Columns = 'COLUMNS',
-}
-
-enum IValueInputOption {
-  USER_ENTERED = 'USER_ENTERED',
-  RAW = 'RAW',
-}
-
 export default class GoogleSheets {
   public spreadsheetId: string;
-  public sheetName: string;
-  public sheetMetadata: any;
-  public headers: any[];
   protected connection: any;
 
-  constructor(spreadsheetId: string, sheetName: string) {
+  constructor(spreadsheetId: string) {
     this.spreadsheetId = spreadsheetId;
-    this.sheetName = sheetName;
+  }
+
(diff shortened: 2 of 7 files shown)


functions/src/services/GoogleSheets.ts
```diff
@@ -197,6 +232,59 @@ export default class GoogleSheets {
...
 
+  /**
+   * Adds a list of data to specific row indices.
+   * First it adds empty rows, then it fills these empty rows
+   * with the provided list of data.
+   * @param rows Array of objects, each holds the data
+   *  to be written into a specific row
+   * @param ranges Array of row indices, where the data should be placed
+   */
+  public async addRows(rows: String[], ranges: number[]): Promise<any> {
```
anantakrishna: I'm not sure this input data structure is good. What if the arrays will have different length?

Maybe we can use [Sparse Arrays](http://2ality.com/2012/06/dense-arrays.html) for that?
https://stackoverflow.com/a/1510842

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,107 @@
...
+    );
+    const onlineSheetRows = await sourceSpreadsheet.getRows();
+
+    const destSpreadsheets = {};
+    destSpreadsheets['Guidelines'] = new GoogleSheets(
+      functions.config().cr.processing.spreadsheet.id,
+      'Guidelines' // can be replaced with any sheet name in the target spreadsheet
+    );
+
+    const sheetNames = await destSpreadsheets['Guidelines'].getSheetNames();
```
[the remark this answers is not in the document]
anantakrishna: If you need to discuss this — please do it in Trello.

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,107 @@
+import * as functions from 'firebase-functions';
+
+import * as helpers from '../helpers';
+import GoogleSheets from '../services/GoogleSheets';
+
+export const copyCRDataToProcessing = functions
```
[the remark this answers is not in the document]
anantakrishna: The function is not renamed yet.

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,96 @@
...
+        const audioFileName = row['Audio File Name'];
+        const submissionSerial = row['Submission Serial'];
+
+
+        if (submissionSerials[sheetName].indexOf(submissionSerial) > -1)
+          continue;
+
+        let index = -1;
+        const sortedFileNames = audioFileNames[sheetName].sort();
+        for (let i = 0; i < sortedFileNames.length; i++) {
```
anantakrishna: Use binary search: https://lodash.com/docs/4.17.11#sortedIndex

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,96 @@
...
+          if (audioFileName < sortedFileNames[i]) {
+            index = i;
+            audioFileNames[sheetName] = [
```
anantakrishna: I understand the reason why you are trying to modify the `audioFileNames` array here, to adjust the index according to your inserted rows. But let's do it in another way.

- Keep original file names array intact.
- Get the original index for insertion by https://lodash.com/docs/4.17.11#sortedIndex.
- Of course, there may be other rows that have already been processed by you and targeted to be inserted before this index.
- In order to keep track of how many rows you are inserting at particular original index use the following construct: `insertions[index]++`.
- To find out the adjusted index for the current row, just sum up all the values in `insertions` array having index less or equal to you original index.

Should work out?

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,107 @@
...
+          list
+        );
+
+        // If the current sheet (list) doesn't exist on the target spreadsheet
+        if (sheetNames.indexOf(list) < 0) {
+          // 1. Create a new sheet with the curent list name
+          await destSpreadsheets[list].createSheet(list);
+
+          // 2. Add the headers of the original sheet (ONLINE sheet)
+          await destSpreadsheets[list].appendRow(sourceSpreadsheet.headers, true);
```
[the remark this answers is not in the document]
anantakrishna: Pay attention to this remark as well.

ahmad-elbatanouni:
Module `ReportingSync` is again not found inside the **modules** folder, however it is imported into the index file.. and this time I'm sure I didn't delete it :-D, I just removed the import

ahmad-elbatanouni:
Also the use of **one time variable** on

```javascript
const insertionSum = insertions[list].reduce((acc, curr) => curr ? curr : 0, 0);
```
is only because the line is too long to be used as an array index.

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,85 @@
...
+    // Holds the rows read from the `Online` sheet
+    // grouped by the sheet (list) name
+    const valuesToAppend = {};
+
+    const insertions = {};
+
+    for (const row of onlineSheetRows) {
+      const list = helpers.extractListFromFilename(row['Audio File Name']);
+
+      if (!destSheets[list]) {
```
anantakrishna: Let's make it systematic. When you meet a first file from some list, initialize everything at once: `destSheets`, `submissionSerials`, `audioFileNames`, `insertions`, `valuesToAppend`.

Moreover, let's use [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of a bunch of plain objects. Here is my proposal:

```
const destination = new Map();
…
if (!destination.has(list)) {
  const sheet = await destSpreadsheet.useSheet(list);
  destination.set(list, {
    sheet,
    submissionSerials: await sheet.getColumn('Submission Serial'),
    audioFileNames: await sheet.getColumn('Audio File Name'),
    insertions: new Map(),
    rows: [] // This is in replace to valuesToAppend
  });
```

Your life will become easier, and the code will become more elegant.

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      if (insertions[list][insIndex])
+        insertions[list][insIndex]++;
+      else
+        insertions[list][insIndex] = 1;
+
+      if (!valuesToAppend[list])
+        valuesToAppend[list] = [];
+
+      const insertionSum = insertions[list].reduce((acc, curr) => curr ? curr : 0, 0);
```
anantakrishna: I believe the logic here is incorrect. You need to sum all the insertions for `index <= originalIndex`. And let's call this variable `indexCorrection`.

functions/src/modules/CR.functions.ts
```diff
@@ -0,0 +1,85 @@
...
+    // to insert a new submission
+    const audioFileNames = {};
+
+    // Holds the rows read from the `Online` sheet
+    // grouped by the sheet (list) name
+    const valuesToAppend = {};
+
+    const insertions = {};
+
+    for (const row of onlineSheetRows) {
```
anantakrishna:
```suggestion
    for (const row of await onlineSheet.getRows()) {
```

functions/src/modules/CR.functions.ts
ahmad-elbatanouni: Thanks, that's much cleaner than what I did

PR #85, opened 2019-03-04
Choosing several languages in the SQR allotment form

DadImScared:


DadImScared:
[Choosing several languages in the SQR allotment form](https://trello.com/c/BHz7UiyO/130-choosing-several-languages-in-the-sqr-allotment-form)

PR #95, opened 2019-03-18
Phone number should not be required in the Donation Receipt Form

DadImScared:


DadImScared:
[Phone number should not be required in the Donation Receipt Form](https://trello.com/c/WbWtoH00/135-phone-number-should-not-be-required-in-the-donation-receipt-form)

PR #96, opened 2019-03-28
Move cr allotment form to the main app

DadImScared:


DadImScared:
[Move CR allotment form to the main app](https://trello.com/c/qucR6yWr/134-move-cr-allotment-form-to-the-main-app)

frontend/src/views/CRAllotment.vue
```diff
@@ -0,0 +1,194 @@
...
+    const spareFiles = await firebase
+      .functions()
+      .httpsCallable("CR-getSpareFiles")(this.filter);
+
+    this.files = spareFiles.data;
+  }, 1000);
+
+  @Watch("filter", { deep: true })
+  handleFilter() {
+    this.debouncedFilter();
```
anantakrishna: What are the benefits of extracting this as a separate function? The function is used only once, and I think it complicates understanding the code. Please consider moving the code inside `handleFilter`.

frontend/src/views/CRAllotment.vue
```diff
@@ -0,0 +1,194 @@
...
+  static initialFilter() {
+    return {
+      languages: [] as string[],
+      list: null
+    };
+  }
+
+  async mounted() {
+    this.getAssignees();
+    this.getLists();
```
anantakrishna: Will these two methods work in parallel? I don't think so because of `await` is used inside them. The result is that lists are loaded only after assignees are loaded. That is why I was using `then` in `SQRAllotment`. Correct me if I'm wrong. If I'm correct then let's bring the code back to this method. I'm not against code splitting, but in this case this method is not that huge, it's easily comprehensible without splitting.

frontend/src/views/CRAllotment.vue
```diff
@@ -0,0 +1,194 @@
...
+    if (this.$route.query.emailAddress) {
+      this.allotment.assignee = this.assignees.find(
+        (assignee: any) =>
+          assignee.emailAddress === this.$route.query.emailAddress
+      );
+    }
+  }
+
+  async getLists() {
+    this.filter.languages = this.languages;
```
anantakrishna: I'm not following this logic. Why do we reset languages in this method?

frontend/src/views/CRAllotment.vue
DadImScared: It appears getLists also fetches data without waiting for assignees.

frontend/src/views/CRAllotment.vue
DadImScared: Debounce returns a function to call.

frontend/src/views/CRAllotment.vue
anantakrishna: ok then.

PR #99, opened 2019-04-06
TE Allotment Processing

forsajan:


forsajan:
[TE Allotment Processing](https://trello.com/c/gG2AebXm/142-te-allotment-processing)

email/track-editing-allotment.html
```diff
@@ -0,0 +1,68 @@
...
+      <tr>
+        <th scope="col">Output File Name</th>
+        <th>Task Definition</th>
+        <th>Unwanted Parts</th>
+      </tr>
+    </thead>
+    <tbody valign="top">
+      {% for task in params.tasks %}
+      <tr>
+        <td>{{ task.outputFileLink }}</td>
```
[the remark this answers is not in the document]
forsajan: Done

email/track-editing-allotment.html
```diff
@@ -26,22 +26,22 @@
     <tbody valign="top">
       {% for task in params.tasks %}
       <tr>
-        <td>{{ task.outputFileLink }}</td>
+        <td>{{ task.fileName }}</td>
         <td>
-          {% for chunk in task.chunks %}
-          {{ chunk.fileName }};&nbsp;{{ chunk.beginning }}-{{ chunk.ending
+          {% for chunk in task.task.chunks %}
+          <a href="{{ params.fileLinkBaseUrl }}/{{ task.listName }}/{{ chunk.fileName }}">{{ chunk.fileName }}</a>&nbsp;{{ chunk.beginning }}&dash;{{ chunk.ending
```
anantakrishna: 1. I would prefer if we construct the links in the function code and inject them into chunks objects rather than putting this logic in the template. So here you should refer them as `chunk.sourceFileLink`.
1. Hypothetically one track editing task can consist of the chunks from different lists. It may be rare or even not be the case at all, only practice will show. But to be safe use each chunk file name to get its list and construct the `sourceFileLink`.
1. `&dash;` is the same as hyphen. You should use `&ndash;` here. See https://dev.w3.org/html5/html-author/charref.

email/track-editing-allotment.html
```diff
@@ -26,22 +26,22 @@
     <tbody valign="top">
       {% for task in params.tasks %}
       <tr>
-        <td>{{ task.outputFileLink }}</td>
+        <td>{{ task.fileName }}</td>
         <td>
-          {% for chunk in task.chunks %}
-          {{ chunk.fileName }};&nbsp;{{ chunk.beginning }}-{{ chunk.ending
+          {% for chunk in task.task.chunks %}
```
anantakrishna: I think this duplication `task.task` can be avoided by injecting `fileName` into the original task obtained from the database.

functions/src/modules/TE.functions.ts
```diff
@@ -37,45 +35,48 @@ export const processAllotment = functions.https.onCall(
...
-    });
+        await teDbRef.update({
+          status: 'Given',
+          assignee: assignee,
+          givenTimestamp: moment().format(),
+        });
+
+        // Getting the tasks list to be used when notifying the assignee
+        const taskRef = await teDbRef.once('value');
+        return { listName: list, fileName: taskId, task: taskRef.val() };
```
anantakrishna: As I mentioned in the template file review, you can inject `fileName` inside the object obtained from `taskRef.val()` and return this object. This way you will avoid `task.task`.

functions/src/modules/TE.functions.ts
```diff
@@ -37,45 +35,48 @@ export const processAllotment = functions.https.onCall(
...
-      allTasks.push(taskRef.val());
-    });
+        await teDbRef.update({
+          status: 'Given',
+          assignee: assignee,
+          givenTimestamp: moment().format(),
+        });
+
+        // Getting the tasks list to be used when notifying the assignee
+        const taskRef = await teDbRef.once('value');
```
anantakrishna: The result of `once` [is called a snapshot](https://firebase.google.com/docs/reference/admin/node/admin.database.Reference#once), so it's better not to introduce a confusion by calling this `taskRef`.

functions/src/modules/TE.functions.ts
```diff
@@ -37,45 +35,48 @@ export const processAllotment = functions.https.onCall(
...
+    const allTasks = await Promise.all(
+      tasks.map(async (taskId: string) => {
+        const list = helpers.extractListFromFilename(taskId);
 
-      await db.ref(`/edited/${list}/${taskId}/trackEditing`).update({
-        status: 'Given',
-        assignee: assignee,
-        givenTimestamp: moment().format('MM/DD/YYYY'),
-      });
+        const teDbRef = db.ref(`/edited/${list}/${taskId}/trackEditing`);
```
anantakrishna: I'd call it `taskRef`.
And one thought stroke me just now — it would be good to test that all the tasks are really found in the database and are spare! I'm adding this to the card description.

PR #103, opened 2019-04-17
Use bolt for database security

DadImScared:


rules.bolt
```diff
@@ -0,0 +1,124 @@
...
+
+type WriteOnlyCoordinator {
+  write() { hasRole("coordinator") }
+}
+
+type OnlyNew {
+  validate() { prior(this) == null }
+}
+
+type EditStatus extends WriteOnlyCoordinator {
```
anantakrishna: I think we can avoid this for now, as various phases will have different statuses. Remove this please.

rules.bolt
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+  }
+}
+
+/original is ReadOnlyCoordinator {
+  /{list} {
+    index() { ["soundQualityReporting/status"] }
+    /{fileName} {
+      /contentReporting {
```
anantakrishna: This whole block repeats below as well. And it is going to be the same for track editing also. Can we extract some common block?

rules.bolt
```diff
@@ -0,0 +1,124 @@
...
+        )
+      }
+      /complete is OnlyNew {}
+      /created is OnlyNew {}
+    }
+  }
+}
+
+/edited/{list}/{fileName} is ReadOnlyCoordinator {
+  /trackEditing {
```
anantakrishna: Here again two blocks are identical. Better to refactor and use common definition.

firebase.json
```diff
@@ -1,6 +1,6 @@
 {
   "database": {
-    "rules": "./database.rules.json"
+    "rules": "rules.bolt"
```
anantakrishna: Shouldn't we add a `firebase-bolt` to the dev dependencies?

DadImScared: I think it's preferred to be a global install.

PR #106, opened 2019-05-04
Fix sqr submission form

DadImScared:


DadImScared:
[SQR submission form](https://trello.com/c/0sdo0Mic/86-sqr-submission-form)

functions/src/modules/SQR.functions.ts
```diff
@@ -648,3 +648,33 @@ export const getSpareFiles = functions.https.onCall(
...
+      const originalFile = file.val();
+      if (!originalFile) {
+        throw new functions.https.HttpsError("invalid-argument", "Invalid token")
+      }
+      const newComments = isCoordinator ?
+          comments
+          :
+          `${originalFile.soundQualityReporting.notes}\n ${comments}`;
+      return file.ref.update({
+        "soundQualityReporting/status": status === 1 ? "Audio Problem" : "",
```
anantakrishna: Don't use magic value `1`. Rather use some string.

functions/src/modules/SQR.functions.ts
```diff
@@ -648,3 +648,33 @@ export const getSpareFiles = functions.https.onCall(
       .slice(0, count || 20);
   }
 );
+
+export const cancelAllotment = functions.https.onCall(
+    async (
+        { fileName, comments, token, status },
```
anantakrishna: Pass `reason` instead of the `status` here. This is function's logic to transform reason into a status change.

functions/src/modules/SQR.functions.ts
```diff
@@ -648,3 +648,33 @@ export const getSpareFiles = functions.https.onCall(
...
+      const isCoordinator = authToken && authToken.coordinator;
+      const listId = fileName.split("-")[0];
+      const file = await admin.database().ref(
+          `original/${listId}/${fileName}`
+      ).orderByChild("token").equalTo(token).once("value");
+      const originalFile = file.val();
+      if (!originalFile) {
+        throw new functions.https.HttpsError("invalid-argument", "Invalid token")
+      }
+      const newComments = isCoordinator ?
```
anantakrishna: Coordinator is not cancelling the allotment this way. So there is no need checking this. Just always add the comments to the notes.

functions/src/modules/SQR.functions.ts
```diff
@@ -648,3 +648,33 @@ export const getSpareFiles = functions.https.onCall(
...
+    async (
+        { fileName, comments, token, status },
+        {
+          auth: {
+            token: authToken
+          } = { token: { coordinator: false } }
+        } = {} as any
+    ) => {
+      const isCoordinator = authToken && authToken.coordinator;
+      const listId = fileName.split("-")[0];
```
anantakrishna: Use `helpers.extractListFromFilename`.

frontend/src/views/SQR/Form.vue
```diff
@@ -280,6 +313,25 @@ export default class Form extends Vue {
     }
   }
 
+  async cancelForm() {
+    const {
+      params: { fileName, token }
+    } = this.$route;
+    const comments = this.cancelComments[this.cancel || 1] || "";
```
anantakrishna: This line looks strange.

frontend/src/views/SQR/Form.vue
```diff
@@ -225,11 +233,13 @@ export default class Form extends Vue {
     }
   ];
   cancel: number | null = null;
-  cancelComments = {};
+  cancelComments: { [key: number]: string } = {};
```
anantakrishna: I'm not sure we need to keep both comments separate. Why not to use the same property for both cases?

frontend/src/views/SQR/Form.vue
```diff
@@ -6,7 +6,10 @@
     <div v-if="isLoadingForm" class="d-flex justify-center">
       <v-progress-circular indeterminate />
     </div>
-    <v-form v-else ref="form" @submit.prevent="handleSubmit">
+    <v-container v-else-if="cancelComplete">
+      <p>Cancel complete!</p>
```
```suggestion
      <p>Allotment is canceled!</p>
```

PR #107, opened 2019-06-18
Auto-saving SQR submission form

DadImScared:


frontend/src/views/SQR/Form.vue
```diff
@@ -242,6 +261,14 @@ export default class Form extends Vue {
   isLoadingForm = true;
   canSubmit = false;
   cancelComplete = false;
+  draftMessages = {
```
anantakrishna: I would call it `formStateMessages`

frontend/src/views/SQR/Form.vue
```diff
@@ -242,6 +261,14 @@ export default class Form extends Vue {
...
+    [FormState.LOADING]: "Saving...",
```
anantakrishna: Why this state is not `FormState.Saving`?

PR #116, opened 2019-07-10
SQR: Make Another Allotment bug

DadImScared:


frontend/src/views/SQRAllotment.vue
```diff
@@ -200,8 +202,8 @@ export default {
       }
     },
     reset() {
-      Object.assign(this.$data.allotment, this.$options.data().allotment);
-      Object.assign(this.$data.filter, this.$options.data().filter);
+      this.filter = CRAllotment.initialFilter();
```
anantakrishna: Why is CR allotment used here, in SQR allotment form? For me it's nonsense. Perhaps the code need re-structuring.

DadImScared: Initially SQRAllotment did not need that method so it was put in CRAllotment as a static method. Maybe it should be be split from CRAllotment so it's more clear?

anantakrishna: Of course, extract common code. It's improper to reference CR form from SQR form.

PR #130, opened 2019-08-11
TE Allotment Form

DadImScared:


frontend/src/components/TE/TaskDefinition.vue
```diff
@@ -0,0 +1,56 @@
...
+  @Prop() item!: any;
+
+  getLink(fileName: string) {
+    const listId = getListId(fileName);
+    return `https://original.${
+      process.env.VUE_APP_STORAGE_ROOT_DOMAIN
+    }/${listId}/${fileName}.flac`;
+  }
+
+  formatSeconds(seconds: number) {
```
anantakrishna: This function is duplicated here and in Unwanted Parts. Consider extracting to utilities.

frontend/src/components/TE/TaskDefinition.vue
```diff
@@ -0,0 +1,56 @@
...
+            >
+              {{chunk.fileName}}
+            </a>
+          </div>
+        </v-flex>
+        <v-flex xs8 sm2 md2 xl1>
+          <span>{{ formatSeconds(chunk.beginning) }}&#8211;{{ formatSeconds(chunk.ending) }}</span>
+        </v-flex>
+        <v-flex sm12 md8 lg8 xl9>
+          <unwanted-parts :item="chunk" />
```
anantakrishna: Why not to pass unwanted parts directly?
```suggestion
          <unwanted-parts :item="chunk.unwantedParts" />
```

frontend/src/components/TE/UnwantedParts.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <div>
+    <p class="font-weight-bold mb-0">{{ item[".key"] }}</p>
+    <ul>
+      <li
+        v-for="(unwantedPart, key, index) in item.unwantedParts"
+        :key="index"
+      >{{ `${formatSeconds(unwantedPart.beginning)}&#8211;${formatSeconds(unwantedPart.ending)},
```
anantakrishna: Use `&ndash;`, it's more self-explanatory.

frontend/src/router.ts
```diff
@@ -136,6 +136,22 @@ export const router = new Router({
             }
           ]
         },
+        {
+          path: "te/",
+          component: () => import("@/views/TE/TE.vue"),
```
anantakrishna: What is the use of this empty component?

DadImScared: It nests the routes.

anantakrishna: I see that it is not TE-specific. Can't we have a common “proxy” component to nest the routes?

DadImScared: I have not been able to find a solution like that because not all nested routes have the same routes in them. Also the navigation menu is generated from the routes.

PR #132, opened 2019-08-21
TE task page

DadImScared:


frontend/public/index.html
```diff
@@ -17,6 +17,7 @@
       href="https://use.fontawesome.com/releases/v5.0.13/css/all.css"
       rel="stylesheet"
     />
+    <link href="https://cdn.materialdesignicons.com/4.0.96/css/materialdesignicons.min.css" rel="stylesheet">
```
anantakrishna: Can't we find all the icons in one single icon set? I wouldn't like us to use two sets.

frontend/src/components/TE/Versions.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <div>
+    <article v-for="(version, index) in versions" :key="index">
+      <h3>Version: {{ index + 1 }}</h3>
+      <p>
+        {{ formatDurationUtc(version.timestamp, "MM-DD-YYYY hh:mm a", "ms") }}
```
anantakrishna: Either the function is called incorrectly or it is used incorrectly. Here we render a timestamp, not a duration. And not in UTC but in the local timezone. Please reconsider.

frontend/src/components/TE/Versions.vue
```diff
@@ -0,0 +1,33 @@
...
+</template>
+
+<script lang="ts">
+import { Component, Prop, Mixins } from "vue-property-decorator";
+import FormatDurationUtc from "@/mixins/FormatDurationUtc";
+
+@Component({
+  name: "Versions"
+})
+export default class Versions extends Mixins<FormatDurationUtc>(
```
anantakrishna: I cannot understand the reasoning for this mix-in. Please make it obvious or reconsider it.

frontend/src/mixins/FormatDurationUtc.ts
```diff
@@ -3,9 +3,13 @@ import moment from "moment";
...
 export default class FormatDurationUtc extends Vue {
-  formatDurationUtc(duration: any, format: string) {
-    return moment
-      .utc(moment.duration(duration, "seconds").asMilliseconds())
-      .format(format);
+  formatDurationUtc(
+    duration: any,
+    format: string,
+    unit: moment.DurationInputArg2 = "seconds",
+    useLocal = false
```
anantakrishna: This flag contradicts the function name - UTC.

frontend/src/mixins/FormatTimeLocal.ts
```diff
@@ -0,0 +1,11 @@
+import { Component, Vue } from "vue-property-decorator";
+import moment from "moment";
+
+@Component
+export default class FormatTimeLocal extends Vue {
+  formatTimeLocal(time: moment.MomentInput, format: string) {
```
anantakrishna: Currently you are using this function only in conjunction with `getDateFormat`, and the calls are looking non-intuitive and redundant. Like this: `formatTimeLocal(task.timestampGiven, getDateFormat(task.timestampGiven))`.

Just merge these two functions into `formatTimestamp` function that will always output the timestamp in the local format, date for past days, time for today.

One more idea: why not to have a single mixin for all timestamp/duration formatting functions? Wouldn't it be more intuitive than having many mixins? This is just a thought, I don't know the architecture and best practices of Vue, so you can refuse this proposal based on your experience.

frontend/src/views/TE/Task.vue
```diff
@@ -0,0 +1,185 @@
...
+          .update({
+            ...this.form,
+            timestamp: firebase.database.ServerValue.TIMESTAMP
+          });
+      }
+    });
+  }
+
+  getDateFormat(timestamp: number) {
+    return moment(timestamp).format("MM-DD-YYYY") ===
```
anantakrishna: https://momentjs.com/docs/#/query/is-same/
> Passing in `day` will check `day`, `month`, and `year`

frontend/src/mixins/FormatDurationUtc.ts
```diff
@@ -3,9 +3,13 @@ import moment from "moment";
 
 @Component
 export default class FormatDurationUtc extends Vue {
-  formatDurationUtc(duration: any, format: string) {
+  formatDurationUtc(
```
anantakrishna: This function is not used in the current branch code. Do you think we should generalize the function as a part of this task? I doubt. Let's put off these changes until they are really needed by other code. So just revert changes of this file.

anantakrishna: Except if you agree to merge mixins (see below), then changes would be justified.

PR #138, opened 2019-08-30
TE Tasks List

DadImScared:


frontend/src/components/DataTable/Link.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <router-link :to="`/te/tasks/${item['.key']}`">{{ item['.key'] }}</router-link>
```
anantakrishna: The component if generic, so you should not use `te` path here. Either pass it as a parameter, or make the component TE-specific.

frontend/src/components/TE/LastVersionMixin.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, Vue } from "vue-property-decorator";
+import _ from "lodash";
+
+@Component
+export default class LastVersionMixin extends Vue {
+  get version() {
+    return (
+      this.item.versions &&
+      _.get(this.item, `versions[${this.item.versions.length - 1}]`, false)
```
anantakrishna: I don't see the justification of using `lodash` here. Can you explain why we can't just use `this.item.versions[this.item.versions.length - 1]`?

frontend/src/components/TE/LastVersionMixin.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, Vue } from "vue-property-decorator";
+import _ from "lodash";
+
+@Component
+export default class LastVersionMixin extends Vue {
+  get version() {
```
anantakrishna: I'd call this property `lastVersion`.

anantakrishna: As well as other properties: `lastResolution` and `isLastApproved`.

frontend/src/components/TE/Output.vue
```diff
@@ -0,0 +1,36 @@
...
+) {
+  @Prop() item!: any;
+  @Prop() value!: string;
+
+  get timestamp() {
+    return this.formatTimestamp(this.version.timestamp);
+  }
+
+  getLink(item: any) {
+    return `https://te.uploads.${process.env.VUE_APP_PROJECT_DOMAIN}/${
```
anantakrishna: If you are constructing this URL in several parts of the code, consider extracting this code to avoid repetitions.

frontend/src/components/TE/TimestampGiven.vue
```diff
@@ -0,0 +1,34 @@
...
+
+<script lang="ts">
+import { Component, Prop, Mixins } from "vue-property-decorator";
+import moment from "moment";
+import FormatTime from "@/mixins/FormatTime";
+
+@Component({
+  name: "TimestampGiven"
+})
+export default class DateGiven extends Mixins<FormatTime>(FormatTime) {
```
anantakrishna: This mixin can be useful outside TE section in other sections, so make it generic (common).
And why the class name differs from the component name?

frontend/src/utility.ts
```diff
@@ -3,6 +3,8 @@ import _ from "lodash";
 import moment from "moment";
 import { ICount } from "@/types/DataTable";
 
+export const teUploadPath = `te.uploads.${process.env.VUE_APP_PROJECT_DOMAIN}`;
```
anantakrishna: It's `trackEditingUploadsBucket`.

PR #149, opened 2019-09-20
Home page buttons should be generated from routes

DadImScared:


frontend/src/components/RequireClaimsButtons.vue
```diff
@@ -1,48 +1,79 @@
...
-import firebase from "firebase/app";
-import "firebase/auth";
+import _ from "lodash";
+import { mapActions } from "vuex";
+import { RouteConfig } from "vue-router";
+import { filterRoutesByClaims, hasClaim } from "@/router";
 
 @Component({
-  name: "RequireClaimsButtons"
+  name: "RequireClaimsButtons",
```
anantakrishna: Please rename this component.
I'd suggest `HomePageButtons`, as this component is going to be used only there. The fact that the buttons are shown according to the claims is secondary. The primary is the purpose of the component—shortcuts on the home page.

frontend/src/components/RequireClaimsButtons.vue
```diff
@@ -1,48 +1,79 @@
...
   }
 
-  get buttons() {
-    return this.initialButtons.filter((button: any) => {
-      return Object.entries(this.userClaims).some(([claimName, claimValue]) => {
-        return button.claims[claimName] === claimValue;
-      });
-    });
-  }
+  makeButtons() {
```
anantakrishna: Personally, I'd move this filtering logic into the router itself. Why not to export something like `getHomePageRoutes` from the router and just transform the routes into the buttons here in this component? I think it will satisfy SoC principle better.

frontend/src/router.ts
```diff
@@ -101,7 +104,10 @@ export const router = new Router({
             {
               path: "tasks",
               component: () => import("@/views/TE/Tasks.vue"),
-              meta: { menuItem: true }
+              meta: {
+                menuItem: true,
+                homePageLink: { text: "Tasks" }
```
anantakrishna:
```suggestion
                homePageLink: { text: "TE" }
```

frontend/src/views/Layout/MainLayout.vue
```diff
@@ -121,7 +121,7 @@ export default class MainLayout extends Vue {
 
   getMenuItems() {
     if (this.userClaims) {
-      return filterRoutesByClaims(this.routes, this.userClaims);
+      return filterRoutesByClaims()(this.routes, this.userClaims);
```
anantakrishna: Here also, let's just export `getMenuItems` from the router and use it here. Let the router filter by claims and meta itself.

frontend/src/router.ts
```diff
@@ -163,23 +163,10 @@ export const hasClaim = (
...
-      userClaims,
-      requireClaims
-    );
-    if (childRoutes.length) return { ...route, children: childRoutes };
-  } else if (requireClaims) {
-    if (hasClaim(requireClaims, userClaims)) return route;
-  } else {
-    return route;
-  }
+export const getRouteChildren = () => {
```
anantakrishna: The purpose of this function is not clear. The children of which route does it return? I cannot understand this from the code.

Moreover, I doubt it should be exported. It is used only twice and in both cases it is used to be passed back to the other router function. Then why not to call it directly from the `getHomePageRoutes` and `getMenuItems` functions?

frontend/src/components/HomePageButtons.vue
```diff
@@ -0,0 +1,55 @@
...
+import { RouteConfig } from "vue-router";
+import { getHomePageRoutes, getRouteChildren } from "@/router";
+
+@Component({
+  name: "HomePageButtons",
+  methods: {
+    ...mapActions("user", ["getUserClaims"])
+  }
+})
+export default class RequireClaimsButtons extends Vue {
```
anantakrishna: Class name should be corrected.

frontend/src/components/HomePageButtons.vue
```diff
@@ -0,0 +1,55 @@
...
+  userClaims: any = {};
+  routeButtons: any[] = [];
+
+  async mounted() {
+    this.userClaims = (await this.getUserClaims()) || {};
+    this.makeButtons(getHomePageRoutes(getRouteChildren(), this.userClaims));
```
anantakrishna: I'm sure `getHomePageRoutes` can itself get current user claims and give all accessible routes for home page. Simplify the function interface.

frontend/src/views/Layout/MainLayout.vue
```diff
@@ -105,23 +105,18 @@ export default class MainLayout extends Vue {
...
-      (route: any) => route.path === "/"
-    ).children;
     this.userClaims = await this.getUserClaims();
   }
 
   get routes(): any {
-    // @ts-ignore
-    return this.navLinks.filter((route: any) => {
+    return getRouteChildren().filter((route: any) => {
       return route.meta && (route.meta.activator || route.meta.menuItem);
```
anantakrishna: This filtering should go to `getMenuItems`. The whole this `routes` property here seems redundant.

frontend/tests/unit/components/HomePageButtons.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+describe("HomePageButtons", () => {
+  let localVue: typeof Vue;
+
+  beforeEach(() => {
+    localVue = createLocalVue();
+    localVue.use(VueRouter);
+  });
+
+  it("should render buttons that match TE claims", () => {
```
anantakrishna: These two test cases are identical except the role. Use Jest's [data-driven tests](https://jestjs.io/docs/en/api#2-testeach-table-name-fn-timeout).

frontend/tests/unit/views/layout/MainLayout.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe("MainLayout", () => {
+  let localVue: typeof Vue;
+
+  beforeEach(() => {
+    localVue = createLocalVue();
+    localVue.use(VueRouter);
+  });
+
+  it("should render menu items based on TE claims", () => {
```
anantakrishna: Here also use data-driven tests.

frontend/src/views/Layout/MainLayout.vue
```diff
@@ -105,23 +105,18 @@ export default class MainLayout extends Vue {
...
     });
   }
 
   getMenuItems() {
```
anantakrishna: This whole function also seems redundant. Just use `getMenuItems` from the router directly.

PR #158, opened 2019-09-28
TE Cancellation

DadImScared:


frontend/src/components/TE/TaskMixin.ts
```diff
@@ -1,8 +1,37 @@
...
+    item: any,
+    itemPath: string,
+    paths: any,
+    fieldUpdates: any
+  ) {
+    try {
+      await firebase.functions().httpsCallable("TE-cancelAllotment")({
+        taskId: item[".key"]
+      });
+      this.updateFields(item, fieldUpdates);
```
anantakrishna: We don't need to update the database from the frontend. It is supposed that all this is done by the function. So you need to call the function only.

frontend/src/views/TE/Task.vue
```diff
@@ -135,6 +137,14 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
     this.isCoordinator = claims.coordinator;
   }
 
+  onCancelClick() {
+    this.cancelAllotment(this.task, "", {}, this.cancelData());
```
anantakrishna: This call looks ugly. Isn't it possible to have this method `cancelAllotment` without any parameters and be directly hooked to button's `@click`? As all the necessary info for cancellation is available in the context. In other words, we are not calling this method with other set of parameters, and are not going to.

DadImScared: Currently the cancel callback is shared between tasks view and task view. Tasks view uses the assignee callback which gets called with those parameters. Should the cancel callback be unique for the separate views?

frontend/src/components/TE/TaskMixin.ts
DadImScared: That function only updates the local state on a successful https function call.

anantakrishna: Ok then

frontend/src/views/TE/Task.vue
anantakrishna: I think in this case we can have them separate

PR #159, opened 2019-09-28
Redirect to the default route in the section

DadImScared:


frontend/src/router.ts
```diff
@@ -305,4 +305,29 @@ export const routerBeforeEach: NavigationGuard = async (to, from, next) => {
   next();
 };
 
-router.beforeEach(routerBeforeEach);
+export const redirectSections: NavigationGuard = async (to, from, next) => {
+  if (_.get(to, "meta.activator", false)) {
```
anantakrishna: For the sake of readability I'd negate this condition and `return next()` here to avoid long `if` statement.

anantakrishna: Also, `false` default value seems redundant, and moreover it confuses, looks like we're comparing `meta.activator` to `false`. I guess getting `undefined` would work as well.

frontend/src/router.ts
```diff
@@ -305,4 +305,29 @@ export const routerBeforeEach: NavigationGuard = async (to, from, next) => {
...
+    const activatorChildren = getRootRouteChildren().find(
+      (route: RouteConfig) => `/${route.path}` === `${to.fullPath}/`
+    ).children;
```
anantakrishna: Can't we just use `to` instead of finding it by path?

> to: Route: the target Route Object being navigated to.

frontend/src/router.ts
```diff
@@ -305,4 +305,29 @@ export const routerBeforeEach: NavigationGuard = async (to, from, next) => {
...
+    const userClaims = await getUserClaims();
```
anantakrishna: By the way, why are we using local function if we have `getUserClaims` action in the `user` store module?

frontend/tests/unit/router.spec.ts
```diff
@@ -1,7 +1,38 @@
...
+  test.each`
+    claims                   | expectedPath
+    ${{ TE: true }}          | ${"/te/my"}
+    ${{ coordinator: true }} | ${"/te/tasks"}
+  `(
+    "should redirect to first available child route that matches claims $claims",
+    async ({ claims, expectedPath }) => {
+      mockClaims(claims);
+      to = {
+        fullPath: "/te",
```
anantakrishna: Move this path to the test data table as well. This way you will be able to test various routes and claims combination.

anantakrishna: By the way, can't we get the whole `to` from the real router by path? Why duplicating here?

frontend/src/router.ts
DadImScared: The `to` object from the route guard does not contain children like the routes on the `router` object.

anantakrishna:
Oh, ok then.

PR #174, opened 2019-10-03
Unwanted Parts is a text in TE

DadImScared:
https://trello.com/c/TtdMnxJJ/206-unwanted-parts-is-a-text-in-te


frontend/src/components/TE/UnwantedParts.vue
```diff
@@ -1,12 +1,13 @@
 <template>
   <div>
     <ul>
-      <template v-for="(unwantedPart, key, index) in unwantedParts">
+      <li>{{ unwantedParts }}</li>
+      <template v-for="(unwantedPart, index) in splitUnwantedParts">
         <li
```
anantakrishna: I don't understand the reasoning for including both original `unwantedParts` as a `li` and the same `unwantedParts` split into lines. Sometimes the description of the unwanted part has line breaks in it, hence splitting by `\n` is not proper as it will break one unwanted part into two bullets.

Moreover, I'm pretty sure we don't need to use `ul` anymore. Perhaps using `<pre>` is appropriate to keep the line breaks in the original text.

And moreover, having said that, I feel that we don't need this component anymore. Displaying unwanted parts has effectively streamlined to `<pre>` tag, which can be used in the appropriate places instead of this component.

PR #189, opened 2019-10-10
TE Cancellation

DadImScared:
https://trello.com/c/7zvWPEnU/204-te-cancellation


DadImScared:
> I see that you have kept `InlineAssignEdit` component. Why? We are not going to use this way of cancellation in any data table. I think it should be removed if you don't have any objection.

There are two views not used currently `/se/tasks` and `/sqr/files` but they use the component. If they are used again will this functionality be removed from them?

anantakrishna:
Yes, it will be removed.
I forgot to mention that in the task. So please remove this component altogether.

PR #204, opened 2019-10-18
Zero-padded hour should be accepted in timing

DadImScared:
https://trello.com/c/Ez9nOEPR/222-zero-padded-hour-should-be-accepted-in-timing

### DIFF
--- a/frontend/src/validation.ts
+++ b/frontend/src/validation.ts
@@ -1,6 +1,6 @@
 export const validateDuration = (v: string) => {
   return v
-    ? v.match("^(\\d[:.])?\\d{1,2}[:.]\\d{1,2}$")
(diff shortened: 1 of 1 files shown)


frontend/src/validation.ts
```diff
@@ -1,6 +1,6 @@
 export const validateDuration = (v: string) => {
   return v
-    ? v.match("^(\\d[:.])?\\d{1,2}[:.]\\d{1,2}$")
+    ? v.match("^(\\d{1,2}[:.])?\\d{1,2}[:.]\\d{1,2}$")
```
anantakrishna:
```suggestion
    ? v.match("^(0?\\d[:.])?\\d{1,2}[:.]\\d{1,2}$")
```

PR #209, opened 2019-10-19
Update Renovate Configuration

DadImScared:


anantakrishna:
Try to use presets whenever possible. See how I used them in the functions section.

Moreover, maybe it's better to use different branch prefix for frontend and functions?

PR #245, opened 2019-11-06
Don't open the SQR submission form if the token is invalid

DadImScared:
https://trello.com/c/lLI19ZXG/232-dont-open-the-sqr-submission-form-if-the-token-is-invalid

### DIFF
--- a/frontend/src/store/modules/user.ts
+++ b/frontend/src/store/modules/user.ts
@@ -18,10 +18,14 @@ export default {
     setCurrentUser: (state: any, user: any) => {
       state.currentUser = user;
       if (!user) {
-        router.push({
-          path: "/login",
-          query: { redirect: router.currentRoute.fullPath }
-        });
+        router
+          .push({
+            path: "/login",
+            query: { redirect: router.currentRoute.fullPath }
+          })
+          .catch(() => {
+            // prevent uncaught promise error
+          });
       }
     }
   },
--- a/frontend/src/views/SQR/Form.vue
+++ b/frontend/src/views/SQR/Form.vue
@@ -321,14 +321,6 @@ export default class Form extends Vue {
     this.cancel = this.cancel === cancelField ? null : cancelField;
   }
 
-  async checkIsCoordinator() {
-    this.isCoordinator = _.get(
-      await this.getUserClaims(),
-      "coordinator",
-      false
-    );
-  }
-
   @Watch("form", { deep: true })
   handleFormErrors() {
     if (this.$refs.form) {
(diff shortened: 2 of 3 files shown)


frontend/src/views/SQR/Form.vue
```diff
@@ -348,6 +356,27 @@ export default class Form extends Vue {
...
+      .database()
+      .ref(`SQR/allotments`)
+      .orderByChild("token")
+      .equalTo(token)
+      .once("value")).val();
+    const sqrStatus = _.get<string>(response, `${fileName}.status`, "");
+    if (!sqrStatus) {
+      this.allotmentError =
+        "This allotment is not valid, please contact coordinator.";
+    } else if (!this.isCoordinator && sqrStatus.toLowerCase() === "done") {
```
anantakrishna: I'm not sure it's worth adding `isCoordinator` property that is used only once in the code. Why not to query the user claims here directly? Or even better, add `isCoordinator` to the `user` store module and map it from there.

frontend/src/views/SQR/Form.vue
```diff
@@ -348,6 +356,27 @@ export default class Form extends Vue {
...
+    const {
+      params: { fileName, token }
+    } = this.$route;
+    const response = (await firebase
+      .database()
+      .ref(`SQR/allotments`)
+      .orderByChild("token")
+      .equalTo(token)
+      .once("value")).val();
+    const sqrStatus = _.get<string>(response, `${fileName}.status`, "");
```
```suggestion
    const allotmentStatus = _.get<string>(response, `${fileName}.status`, "");
```

frontend/src/views/SQR/Form.vue
```diff
@@ -381,8 +410,11 @@ export default class Form extends Vue {
   }
 
   async mounted() {
-    this.canSubmit = true;
-    this.getSavedData();
+    await this.checkIsCoordinator();
+    await this.canSubmitForm();
```
anantakrishna: Usually the methods should be named in the form “Do this and that”. `canSubmitForm` looks more like a calculated property. Consider renaming or refactoring this.

frontend/src/store/modules/user.ts
```diff
@@ -18,10 +18,14 @@ export default {
...
-          query: { redirect: router.currentRoute.fullPath }
-        });
+        router
+          .push({
+            path: "/login",
+            query: { redirect: router.currentRoute.fullPath }
+          })
+          .catch(() => {
+            // prevent uncaught promise error
+          });
```
anantakrishna: From the perspective of the task, this change looks irrelevant. Is it for fixing http://prntscr.com/prg5qm?

frontend/src/views/SQR/Form.vue
DadImScared: Where would the initial `isCoordinator` be set in the vuex store? Maybe the vuex commit that runs on [auth state change](https://github.com/PureBhaktiArchive/audioseva/blob/master/frontend/src/main.ts#L59) can be replaced with an action that sets the currentUser and does additional stuff like setting isCoordinator?

frontend/src/store/modules/user.ts
DadImScared: Yes

frontend/src/views/SQR/Form.vue
anantakrishna: ok, let's leave it as of now, but please consider refactoring this whenever you have time. I have added it as a point to https://trello.com/c/vKAD3ViL/17-load-user-info-from-the-routes.

frontend/src/views/SQR/Form.vue
```diff
@@ -312,6 +310,8 @@ export default class Form extends Vue {
   };
   submitSuccess = false;
   formHasError = false;
+  isCoordinator = false;
+  allotmentError = "";
```
anantakrishna: As this property is used not only for the allotment errors, I'd name it more generic. For example,
```suggestion
  errorMessage = "";
```

PR #262, opened 2019-11-22
App is not loading on iOS 8, 9

DadImScared:
https://trello.com/c/Zf1blj0w/246-app-is-not-loading-on-ios-8-9

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,7 +25,8 @@
     "vue2-dropzone": "3.6.0",
     "vuefire": "1.4.5",
     "vuetify": "1.5.18",
-    "vuex": "3.1.1"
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -25,7 +25,8 @@
     "vue2-dropzone": "3.6.0",
     "vuefire": "1.4.5",
     "vuetify": "1.5.18",
-    "vuex": "3.1.1"
+    "vuex": "3.1.1",
+    "whatwg-fetch": "^3.0.0"
```
anantakrishna: Pin the dependency version in order to avoid Renovate doing this afterwards.
```suggestion
    "whatwg-fetch": "3.0.0"
```

PR #264, opened 2019-11-23
Manage page title

DadImScared:
https://trello.com/c/BfR0qG7s/248-manage-page-title

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -20,6 +20,7 @@
     "moment-timezone": "0.5.27",
     "vue": "2.6.10",
     "vue-class-component": "7.1.0",
+    "vue-page-title": "1.1.5",
     "vue-property-decorator": "8.3.0",
     "vue-resource": "1.5.1",
     "vue-router": "3.1.3",
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -19,9 +19,20 @@ import vuetifyOptions from "@/vuetifyOptions";
 import "vuetify/dist/vuetify.min.css";
 
 import { rtdbPlugin as VueFire } from "vuefire";
+import VuePageTitle from "vue-page-title";
 
 import "@babel/polyfill";
 
+Vue.use(VuePageTitle, {
+  suffix: "| Audio Seva",
+  setTitleMethod: (title: string) => {
+    if (title.includes("Home |")) {
+      document.title = "Audio Seva";
+    } else {
+      document.title = title;
+    }
+  }
+});
 Vue.use(VueResource);
 Vue.use(VueFire);
 Vue.use(Vuetify);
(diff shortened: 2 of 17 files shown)


frontend/src/views/Home.vue
```diff
@@ -14,6 +14,7 @@
 import HomePageButtons from "../components/HomePageButtons";
 export default {
   name: "Home",
+  title: "Home",
```
anantakrishna: For home page there is no need in this. Let it be just “Audio Seva”. Possible?

frontend/src/views/SQRAllotment.vue
```diff
@@ -168,6 +168,7 @@ import ErrorMessages from "../mixins/ErrorMessages";
 export default {
   mixins: [ErrorMessages],
   name: "SQRAllotment",
+  title: "Sound Quality Reporting",
```
anantakrishna:
```suggestion
  title: "SQR Allotment",
```

frontend/package.json
```diff
@@ -20,6 +20,7 @@
     "moment-timezone": "0.5.27",
     "vue": "2.6.10",
     "vue-class-component": "7.1.0",
+    "vue-page-title": "^1.1.5",
```
anantakrishna: Pin the version. I believe it is possible to configure `npm` to always add a fixed version instead of a range during install. Don't forget to update the lock file.

frontend/src/views/Home.vue
DadImScared: I think vue-page-title does not support a change like that. I think because Home is the exception a solution might be to change it through `document.title` on the Home page

frontend/src/main.ts
```diff
@@ -19,9 +19,20 @@ import vuetifyOptions from "@/vuetifyOptions";
...
 import "@babel/polyfill";
 
+Vue.use(VuePageTitle, {
+  suffix: "| Audio Seva",
+  setTitleMethod: (title: string) => {
+    if (title.includes("Home |")) {
+      document.title = "Audio Seva";
+    } else {
+      document.title = title;
+    }
```
anantakrishna: Here I'd assume that the title is empty/null rather than `Home |`. You can compare the title to the suffix and if they match then trim the pipe from the suffix. This way you'd avoid duplication of `Home` and of `Audio Seva`.

DadImScared: If no title is used in Home the function is never called.

anantakrishna: That means that the page title will remain the same as on the previous route?

By the way, what do you mean by “no title is used”? For `undefined` it may be true, but what about `null` or empty?

DadImScared: Yes the title stays the same as the previous page. This behavior is there if title is undefined, null or empty.

PR #272, opened 2019-11-26
Get project root domain from URL

DadImScared:
https://trello.com/c/nrImnCCS/247-get-project-root-domain-from-url


frontend/src/utility.ts
```diff
@@ -3,9 +3,14 @@ import _ from "lodash";
...
-  process.env.VUE_APP_PROJECT_DOMAIN
-}`;
+export const getProjectDomain = () => {
+  return window.location.host
+    .split(".")
+    .slice(1)
+    .join(".");
+};
+
+export const trackEditingUploadsBucket = `te.uploads.${getProjectDomain()}`;
```
anantakrishna: I propose to eliminate this const and compile the TE uploads bucket name in the appropriate component. It is used only once.

PR #279, opened 2019-11-29
Uniform layout for all the pages

DadImScared:
https://trello.com/c/ttuQdsaO/251-uniform-layout-for-all-the-pages


frontend/src/router.ts
```diff
@@ -15,54 +15,34 @@ export const router = new Router({
   routes: [
     { path: "*", redirect: "/" },
     {
-      path: "/login",
-      meta: { auth: { guestOnly: true } },
+      path: "/",
+      meta: { auth: { requireAuth: true } },
```
anantakrishna: Do we really need to apply this constraint to the root route? I feel it's redundant because the nested routes which require authentication, they are explicitly requiring a role.

Consider removing this constraint and redundant `requireAuth: false` for anonymous nested routes.

frontend/src/views/Layout/MainLayout.vue
```diff
@@ -1,95 +1,105 @@
...
-import BaseLayout from "./BaseLayout.vue";
 import MenuLinks from "@/components/MenuLinks";
 
+const pathMap = {
+  cr: "CR",
+  sqr: "SQR",
+  te: "TE",
+  allot: "Allotment"
+};
+
```
anantakrishna: I think this knowledge is not relevant here and could be extracted to the routes or to the components. Please consider refactoring this.

PR #282, opened 2019-12-01
TE uploaded files should be downloadable

DadImScared:
https://trello.com/c/bVlBUaCu/257-link-uploaded-te-files-to-the-firebase-storage


frontend/src/components/TE/VersionDownloadLink.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <a v-if="downloadLink" :href="downloadLink" v-bind="$attrs"><slot></slot></a>
```
anantakrishna: What is `v-bind="$attrs"` for?

DadImScared: That allows any attribute that's not a prop to be passed to the `a` tag through the component.

PR #284, opened 2019-12-07
Completed SQR submission in the drafts branch

DadImScared:
https://trello.com/c/YXtRA5aM/259-completed-sqr-submission-in-the-drafts-branch


frontend/src/views/SQR/Form.vue
```diff
@@ -500,6 +497,13 @@ export default class Form extends Vue {
   get isCancelChecked() {
     return Object.values(this.cancelCheck).includes(true);
   }
+
+  get isSubmission() {
```
```suggestion
  get isCompleted() {
```

frontend/src/views/SQR/Form.vue
```diff
@@ -321,6 +318,7 @@ export default class Form extends Vue {
         this.initialData[".value"] !== null ||
         branch === SubmissionsBranch.DRAFTS
       ) {
+        this.branch = branch;
```
anantakrishna: Seems like this is the only place where `branch` is set. What about the following scenario:

- The form is loaded from the `drafts` branch.
- The form is submitted by the user.
- From now on the `branch` should be `completed`, not `drafts`.

Is it covered?

DadImScared: After the form is submitted a different screen is shown and to get back to the form a page refresh is needed so there's no need to assign `this.branch` to `completed`.

frontend/src/views/SQR/Form.vue
```diff
@@ -197,6 +197,7 @@ export default class Form extends Vue {
   showValidationSummary = false;
   getUserClaims!: any;
   rules = [required];
+  branch: SubmissionsBranch | "" = "";
```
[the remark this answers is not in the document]
DadImScared: The data has to have a value to be reactive and I did not want to set a default branch. I have changed it to `null` instead of `""` because I think that makes more sense.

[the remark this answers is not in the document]
anantakrishna: OK.

frontend/src/views/SQR/Form.vue
anantakrishna: I see.

frontend/src/views/SQR/Form.vue
anantakrishna: So rename this and update the pull request.

PR #292, opened 2019-12-11
Fix task background color is wrong color for WIP status

DadImScared:
https://trello.com/c/2Ud97j87/145-te-tasks-list

### DIFF
--- a/frontend/src/components/TE/TaskMixin.ts
+++ b/frontend/src/components/TE/TaskMixin.ts
@@ -1,8 +1,8 @@
-import { Component, Vue } from "vue-property-decorator";
-import _ from "lodash";
+import { Component, Mixins } from "vue-property-decorator";
+import BaseTaskMixin from "@/components/TE/BaseTaskMixin";
 
 @Component
-export default class TaskMixin extends Vue {
+export default class TaskMixin extends Mixins<BaseTaskMixin>(BaseTaskMixin) {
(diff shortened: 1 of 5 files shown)


frontend/src/components/TE/TaskMixin.ts
```diff
@@ -21,8 +21,7 @@ export default class TaskMixin extends Vue {
   getTaskStyle(task: any) {
     let backgroundColor = "inherit";
     const resolution =
-      task.versions &&
-      _.get(task, `versions[${task.versions.length - 1}].resolution`, false);
+      task.versions && (Object.values(task.versions).pop() as any).resolution;
```
anantakrishna: I remember we have `LastVersionMixin`. Why not to use it here?
And can you by the way inject the version key to the returned `lastVersion` in order to use it in the `Output` component's download link?

DadImScared: `LastVersionMixin` gets the task from the data and this accepts the task as an parameter in the function. Maybe a mixin can be added that they both inherit from that has shared methods? To inject the key should it be added into the object that is returned from `lastVersion` computed property?

anantakrishna: I think we can add a common ancestor mixin for this.

And yes, the key should be a property of the returned object from `lastVersion`, name it `id`.

PR #300, opened 2019-12-18
TE task output file should be prompted to save, not opened in the browser

DadImScared:
https://trello.com/c/EONmEN74/266-te-task-output-file-should-be-prompted-to-save-not-opened-in-the-browser

### DIFF
--- a/frontend/src/views/TE/Upload.vue
+++ b/frontend/src/views/TE/Upload.vue
@@ -223,7 +223,7 @@ export default class Upload extends Vue {
     const uploadTask = this.uploadsBucket
       .ref()
       .child(path)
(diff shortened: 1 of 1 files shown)


frontend/src/views/TE/Upload.vue
```diff
@@ -223,7 +223,7 @@ export default class Upload extends Vue {
     const uploadTask = this.uploadsBucket
       .ref()
       .child(path)
-      .put(file);
+      .put(file, { contentDisposition: `attachment; filename=${file.name}` });
```
anantakrishna: Try to set `null` to this metadata field. That would be preferable.

PR #304, opened 2019-12-19
Show Srila Gurudeva's quotes on the loading screen

DadImScared:
https://trello.com/c/cc9Xxdt8/285-srila-gurudevas-quotes-on-the-loading-screen


frontend/src/App.vue
```diff
@@ -3,9 +3,28 @@
...
+  mounted() {
+    this.getQuotes();
+  },
+  methods: {
+    async getQuotes() {
+      await this.$rtdbBind("quotes", firebase.database().ref("quotes"));
+      localStorage.setItem(
+        "quotes",
+        JSON.stringify([
+          "I am closer to you than the air you breathe.",
```
anantakrishna: Don't duplicate this default quote here, only use ones fetched from the database.

frontend/public/index.html
```diff
@@ -29,12 +29,27 @@
     </noscript>
     <div>
       <div id="app">
-        <div class="overlay">
-          <div class="loadingWrapper">
-            <div class="loader">Loading...</div>
```
anantakrishna: There still should be some loading indicator.

frontend/public/index.html
```diff
@@ -29,12 +29,27 @@
...
+        if (quotes) {
+          showQuote(getRandomQuote(JSON.parse(quotes)));
+        } else {
+          let defaultQuote = localStorage.getItem("defaultQuote");
+          if (!defaultQuote) {
+            localStorage.setItem("defaultQuote", "I am closer to you than the air you breathe.");
+            defaultQuote = "I am closer to you than the air you breathe.";
+          }
+          showQuote(defaultQuote);
+        }
```
anantakrishna: I think it's over complicated.

- No need to save the default quote into the storage.
- The text of the default quote should not be mentioned twice in the source code.
- You can pick randomly from one default quote to streamline the code flow.
- You even can save the default quote to the storage if it's empty, later the default quote will be overwritten from the database.

PR #305, opened 2019-12-20
Listening page improvements in IE11

DadImScared:
https://trello.com/c/WuXLXpND/262-listening-page-improvements-in-ie11

### DIFF
--- a/frontend/src/views/Listen.vue
+++ b/frontend/src/views/Listen.vue
@@ -4,7 +4,7 @@
     <div class="pb-3">
       <h3>{{ $title }}</h3>
     </div>
-    <v-card color="#d9edf7" :max-width="isInternetExplorer ? '90vw' : '100%'">
+    <v-card color="#d9edf7">
       <v-card-title primary-title>
         <div>
           <h3 class="headline mb-0">Audio Player</h3>
@@ -53,7 +53,13 @@
   </div>
 </template>
 
-<style></style>
+<style>
+@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+  .v-card {
+    max-width: 90vw;
+  }
+}
+</style>
 
 <script lang="ts">
 import { Component, Vue } from "vue-property-decorator";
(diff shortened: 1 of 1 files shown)


frontend/src/views/Listen.vue
```diff
@@ -4,13 +4,17 @@
...
       <v-card-title primary-title>
         <div>
           <h3 class="headline mb-0">Audio Player</h3>
         </div>
       </v-card-title>
-      <v-card-title primary-title>
+      <v-alert v-if="isInternetExplorer" :value="true">
+        FLAC files are not supported in Internet Explorer. Please download the
+        file.
+      </v-alert>
```
anantakrishna: Maybe we can use https://www.npmjs.com/package/caniuse-lite for this, not hard-coding IE here?

frontend/src/views/Listen.vue
```diff
@@ -4,13 +4,17 @@
     <div class="pb-3">
       <h3>{{ $title }}</h3>
     </div>
-    <v-card color="#d9edf7">
+    <v-card color="#d9edf7" :max-width="isInternetExplorer ? '90vw' : '100%'">
```
anantakrishna: Can we use a pure CSS browser targeting instead like suggested here https://stackoverflow.com/a/28419263/3082178?

frontend/src/views/Listen.vue
DadImScared: I think this needs to be paired with the browser and version of the browser to give accurate results. Maybe we can use https://github.com/lancedikson/bowser to get the browser?

I think we might be able to use https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType even though some browsers that support it return an empty string the browsers that we are aiming to support all return the expected result from `canPlayType` https://caniuse.com/#feat=flac

anantakrishna: Let's use `canPlayType`, it's more suitable for us than pulling another dependency.

PR #307, opened 2019-12-20
Show error message if TE task is not accessible to the user

DadImScared:
https://trello.com/c/IDkOo1Kg/286-show-error-message-if-te-task-is-not-accessible-to-the-user

### DIFF
--- a/frontend/src/views/TE/Task.vue
+++ b/frontend/src/views/TE/Task.vue
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <div v-if="isFetchingTask">
+    <div v-if="taskState === TaskState.LOADING">
       <v-progress-circular indeterminate></v-progress-circular>
     </div>
-    <div v-else-if="fetchTaskError">
+    <div v-else-if="taskState === TaskState.ERROR">
       <v-alert color="warning" :value="true">
         This task is not available to you. Please contact your coordinator.
       </v-alert>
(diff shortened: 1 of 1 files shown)


frontend/src/views/TE/Task.vue
```diff
@@ -187,6 +192,7 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
   task: any = {};
   isFetchingTask = true;
   isCoordinator = false;
+  fetchTaskError = false;
```
anantakrishna: Please read https://itnext.io/managing-local-state-in-vue-js-components-d5d92fb432 and decide which of two approaches is suitable for this component. I opt for “String” approach as the states are mutually exclusive.

frontend/src/views/TE/Task.vue
```diff
@@ -174,6 +174,12 @@ import TaskMixin from "@/components/TE/TaskMixin";
 import FormatTime from "@/mixins/FormatTime";
 import VersionDownloadLink from "@/components/TE/VersionDownloadLink.vue";
 
+enum TaskState {
```
anantakrishna: I'd say it's not a task state but the page state. Did other pages there will be also state enum, and no need to prefix it with a specific term. There will be usually only one enum per page.

frontend/src/views/TE/Task.vue
```diff
@@ -190,9 +196,9 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
   FormatTime
 ) {
   task: any = {};
-  isFetchingTask = true;
+  TaskState = TaskState;
```
anantakrishna: Can we avoid this duplication?

DadImScared: To use the enum in the template I think it needs to be declared in the class.

PR #312, opened 2019-12-22
Ability for the TE coordinator to upload files for any task

DadImScared:
https://trello.com/c/U1u38ycU/309-ability-for-the-coordinator-to-upload-output-te-files-for-any-task

### DIFF
--- a/frontend/src/router.ts
+++ b/frontend/src/router.ts
@@ -93,7 +93,7 @@ export const router = new Router({
               component: () => import("@/views/TE/Upload.vue"),
               meta: {
                 menuItem: true,
-                auth: { requireClaims: { TE: true } },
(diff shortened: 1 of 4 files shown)


frontend/src/router.ts
```diff
@@ -93,7 +93,7 @@ export const router = new Router({
               component: () => import("@/views/TE/Upload.vue"),
               meta: {
                 menuItem: true,
-                auth: { requireClaims: { TE: true } },
+                auth: { requireClaims: { TE: true, coordinator: true } },
```
anantakrishna: Doesn't this require both? It should require any.

DadImScared: The current way to check for claims requires any not both.

PR #320, opened 2020-01-02
Separate roles for coordinators

DadImScared:
https://trello.com/c/gmPD9JBI/138-separate-roles-for-coordinators

### DIFF
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -5,7 +5,6 @@ export const subjects = {
   TE: {
     tasks: "TE/Tasks",
     myTasks: "TE/MyTasks",
-    allot: "TE/Allot",
     task: "TE/Task"
   },
   SQR: {
@@ -39,33 +38,35 @@ const hasRole = (roles: any) => {
 const handleTEPermissions = (can: any, cannot: any) => {
   if (hasRole("TE.coordinator")) {
     can("manage", subjects.TE.task);
-    can("manage", subjects.TE.allot);
     can("manage", subjects.TE.tasks);
-    return;
   }
   if (hasRole("TE.checker")) {
-    can("resolve", subjects.TE.task);
+    can(["resolve", "view"], subjects.TE.task);
+    can("view", subjects.TE.tasks);
   }
   if (hasRole("TE.editor")) {
-    can("upload", subjects.TE.task);
+    can(["upload", "view"], subjects.TE.task);
+    can("view", subjects.TE.myTasks);
   }
 };
 
 const handleSQRPermissions = (can: any, cannot: any) => {
   can("save", subjects.SQR.form, { isCompleted: false });
+
+  if (!hasRole(["SQR.coordinator", "SQR.checker"])) {
+    can("submit", subjects.SQR.form, { isCompleted: false });
+    can("read", subjects.SQR.form, { done: false });
+    return;
+  }
+
   if (hasRole("SQR.coordinator")) {
     can(["submit", "read"], subjects.SQR.form);
     can("allot", subjects.SQR.tasks);
-    return;
   }
 
   if (hasRole("SQR.checker")) {
     can(["submit", "read"], subjects.SQR.form);
-    return;
   }
-
-  can("submit", subjects.SQR.form, { isCompleted: false });
-  can("read", subjects.SQR.form, { done: false });
 };
 
 export const defineAbilities = () => {
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -57,8 +57,8 @@ const handleSQRPermissions = (can: any, cannot: any) => {
   }
 
   if (hasRole("SQR.coordinator")) {
-    can("submit", subjects.SQR.form);
-    can("allot", subjects.SQR.tasks);
+    can("manage", subjects.SQR.form);
+    can("manage", subjects.SQR.tasks);
   }
 
   if (hasRole("SQR.checker")) {
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -1,7 +1,13 @@
 import { AbilityBuilder } from "@casl/ability";
 import { store } from "@/store";
 
-const TESubjects = ["TE/Task", "TE/Upload"];
+const TESubjects = [
+  "TE/Task",
+  "TE/Upload",
+  "TE/Allot",
+  "TE/Tasks",
+  "TE/MyTasks"
+];
 const SQRSubjects = ["SQR/Allot", "SQR/UpdateDone"];
 
 const subjects = [...TESubjects, ...SQRSubjects].reduce(
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -1,0 +1,39 @@
+import { AbilityBuilder } from "@casl/ability";
+import { store } from "@/store";
+
+const tePermissions = ["TE/Task", "TE/Upload"];
+const SQRPermissions = ["SQR/Allot", "SQR/Form"];
+
+const hasRole = (roles: any) => {
+  return store.getters["user/hasRole"](roles);
+};
+
+const handleTEPermissions = (can: any) => {
+  if (hasRole("TE.coordinator")) {
+    can("manage", tePermissions);
+    return;
+  }
+  if (hasRole("TE.checker")) {
+    can("update", tePermissions[0]);
+  }
+  if (hasRole("TE.editor")) {
+    can("read", tePermissions[1]);
+  }
+};
+
+const handleSQRPermissions = (can: any) => {
+  if (hasRole("SQR.coordinator")) {
+    can("manage", SQRPermissions);
+    return;
+  }
+  if (hasRole("SQR.checker")) {
+    can("update", SQRPermissions[1]);
+  }
+};
+
+export const defineAbilities = () => {
+  const { rules, can } = AbilityBuilder.extract();
+  handleTEPermissions(can);
+  handleSQRPermissions(can);
+  return rules;
+};
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -19,7 +19,7 @@ class Data {
   }
 }
 
-export class SQRForm extends Data {
+export class SQRFormAbilityData extends Data {
   static get modelName() {
     return subjects.SQR.form;
   }
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -1,53 +1,76 @@
 import { AbilityBuilder } from "@casl/ability";
 import { store } from "@/store";
 
-const TESubjects = [
-  "TE/Task",
-  "TE/Upload",
-  "TE/Allot",
-  "TE/Tasks",
-  "TE/MyTasks"
-];
-const SQRSubjects = ["SQR/Allot", "SQR/UpdateDone"];
-
-const subjects = [...TESubjects, ...SQRSubjects].reduce(
-  (permissions, permission) => {
-    permissions[permission] = permission;
-    return permissions;
+export const subjects = {
+  TE: {
+    tasks: "TE/Tasks",
+    myTasks: "TE/MyTasks",
+    allot: "TE/Allot",
+    task: "TE/Task"
   },
-  {} as { [key: string]: string }
-);
+  SQR: {
+    tasks: "SQR/Tasks",
+    form: "SQR/Form"
+  }
+};
+
+class Data {
+  constructor(props: any) {
+    Object.assign(this, props);
+  }
+}
+
+export class SQRForm extends Data {
+  static get modelName() {
+    return subjects.SQR.form;
+  }
+}
+
+export const SubjectsPlugin = {
+  install(Vue: any) {
+    Vue.prototype.$subjects = subjects;
+  }
+};
 
 const hasRole = (roles: any) => {
   return store.getters["user/hasRole"](roles);
 };
 
-const handleTEPermissions = (can: any) => {
+const handleTEPermissions = (can: any, cannot: any) => {
   if (hasRole("TE.coordinator")) {
-    can("manage", TESubjects);
+    can("manage", subjects.TE.task);
+    can("manage", subjects.TE.allot);
+    can("manage", subjects.TE.tasks);
     return;
   }
   if (hasRole("TE.checker")) {
-    can("resolve", subjects["TE/Task"]);
+    can("resolve", subjects.TE.task);
   }
   if (hasRole("TE.editor")) {
-    can("upload", subjects["TE/Upload"]);
+    can("upload", subjects.TE.task);
   }
 };
 
-const handleSQRPermissions = (can: any) => {
+const handleSQRPermissions = (can: any, cannot: any) => {
+  can("save", subjects.SQR.form, { isCompleted: false });
   if (hasRole("SQR.coordinator")) {
-    can("manage", SQRSubjects);
+    can(["submit", "read"], subjects.SQR.form);
+    can("allot", subjects.SQR.tasks);
     return;
   }
+
   if (hasRole("SQR.checker")) {
-    can("submit", subjects["SQR/UpdateDone"]);
+    can(["submit", "read"], subjects.SQR.form);
+    return;
   }
+
+  can("submit", subjects.SQR.form, { isCompleted: false });
+  can("read", subjects.SQR.form, { done: false });
 };
 
 export const defineAbilities = () => {
-  const { rules, can } = AbilityBuilder.extract();
-  handleTEPermissions(can);
-  handleSQRPermissions(can);
+  const { rules, can, cannot } = AbilityBuilder.extract();
+  handleTEPermissions(can, cannot);
+  handleSQRPermissions(can, cannot);
   return rules;
 };
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -1,8 +1,16 @@
 import { AbilityBuilder } from "@casl/ability";
 import { store } from "@/store";
 
-const tePermissions = ["TE/Task", "TE/Upload"];
-const SQRPermissions = ["SQR/Allot", "SQR/Form"];
+const TESubjects = ["TE/Task", "TE/Upload"];
+const SQRSubjects = ["SQR/Allot", "SQR/UpdateDone"];
+
+const subjects = [...TESubjects, ...SQRSubjects].reduce(
+  (permissions, permission) => {
+    permissions[permission] = permission;
+    return permissions;
+  },
+  {} as { [key: string]: string }
+);
 
 const hasRole = (roles: any) => {
   return store.getters["user/hasRole"](roles);
@@ -10,24 +18,24 @@ const hasRole = (roles: any) => {
 
 const handleTEPermissions = (can: any) => {
   if (hasRole("TE.coordinator")) {
-    can("manage", tePermissions);
+    can("manage", TESubjects);
     return;
   }
   if (hasRole("TE.checker")) {
-    can("update", tePermissions[0]);
+    can("resolve", subjects["TE/Task"]);
   }
   if (hasRole("TE.editor")) {
-    can("read", tePermissions[1]);
+    can("upload", subjects["TE/Upload"]);
   }
 };
 
 const handleSQRPermissions = (can: any) => {
   if (hasRole("SQR.coordinator")) {
-    can("manage", SQRPermissions);
+    can("manage", SQRSubjects);
     return;
   }
   if (hasRole("SQR.checker")) {
-    can("update", SQRPermissions[1]);
+    can("submit", subjects["SQR/UpdateDone"]);
   }
 };
 
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -39,7 +39,7 @@ const handleTEPermissions = (can: any, cannot: any) => {
 };
 
 const handleSQRPermissions = (can: any, cannot: any) => {
-  can("submit", subjects.SQR.form, { isCompleted: false });
+  can("submit", subjects.SQR.form, { isMarkedDone: false });
 
   if (hasRole("SQR.coordinator")) {
     can("manage", subjects.SQR.form);
--- a/frontend/src/abilities.ts
+++ b/frontend/src/abilities.ts
@@ -13,18 +13,6 @@ export const subjects = {
   }
 };
 
-class Data {
-  constructor(props: any) {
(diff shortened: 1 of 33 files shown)


frontend/src/views/TE/Task.vue
```diff
@@ -206,9 +207,9 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
   currentUser!: firebase.User;
   getUserClaims!: any;
 
-  mounted() {
+  async mounted() {
     this.getTask();
-    this.checkUserClaims();
+    this.claims = await this.getUserClaims();
```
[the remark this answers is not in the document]
anantakrishna: And not claims but roles. However in the light of my proposal to use https://github.com/stalniy/casl it might be completely rethought.

frontend/src/views/TE/Task.vue
```diff
@@ -275,6 +271,18 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
     });
   }
 
+  get canApprove() {
```
anantakrishna: This permission is not just "Approve", but "Check" in general.

frontend/src/views/TE/Task.vue
```diff
@@ -275,6 +271,18 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
...
+    return hasClaim(["TE.checker"], this.claims);
+  }
+
+  get showUpload() {
```
anantakrishna: The permission is "Can Upload".

frontend/tests/unit/HomePageButtons.spec.ts
```diff
@@ -16,8 +16,8 @@ describe("HomePageButtons", () => {
 
   test.each`
     claims
-    ${{ TE: true }}
-    ${{ coordinator: true }}
+    ${{ TE: { editor: true } }}
+    ${{ TE: { coordinator: true } }}
```
anantakrishna: I think the claims will contain just

`roles: ["role1", "role2"]`

frontend/src/router.ts
```diff
@@ -46,7 +46,7 @@ export const router = new Router({
             activator: true,
             activatorName: "Sound Quality Reporting",
             menuIcon: "fas fa-headphones",
-            auth: { requireClaims: { coordinator: true } }
+            auth: { requireClaims: ["SQR.coordinator"] }
```
anantakrishna: Let's replace `requireCLaims` router metadata key with `roles` key for specifying roles for which the route is accessible. Also we can support both `role: string` and `roles: string[]` for convenience.

frontend/src/abilities.ts
```diff
@@ -0,0 +1,39 @@
...
+    can("update", tePermissions[0]);
+  }
+  if (hasRole("TE.editor")) {
+    can("read", tePermissions[1]);
+  }
+};
+
+const handleSQRPermissions = (can: any) => {
+  if (hasRole("SQR.coordinator")) {
+    can("manage", SQRPermissions);
```
anantakrishna: I think the concept here is wrong. It is not a permission, it's a subject of the `manage` action. Consider renaming.

frontend/src/abilities.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  if (hasRole("SQR.checker")) {
+    can("update", SQRPermissions[1]);
+  }
+};
+
+export const defineAbilities = () => {
+  const { rules, can } = AbilityBuilder.extract();
+  handleTEPermissions(can);
+  handleSQRPermissions(can);
```
anantakrishna: Please read https://stalniy.github.io/casl/abilities/roles/2017/07/21/roles.html. I feel we can learn from there in regard to handling roles.

frontend/src/router.ts
```diff
@@ -62,38 +62,38 @@ export const router = new Router({
...
+              meta: { menuItem: true, auth: { roles: ["TE.coordinator"] } }
             },
             {
               path: "my",
               component: () => import("@/views/TE/MyTasks.vue"),
               meta: {
                 menuItem: true,
                 menuLinkName: "My Tasks",
-                auth: { requireClaims: { TE: true } }
+                auth: { roles: ["TE.editor", "TE.coordinator"] }
```
anantakrishna: Coordinator does not need access to My Tasks.

frontend/src/router.ts
```diff
@@ -233,24 +201,24 @@ export const checkAuth: NavigationGuard = async (to, from, next) => {
...
 
-  if (requireClaims) {
+  if (roles) {
     const userClaims = currentUser
-      ? (await currentUser.getIdTokenResult()).claims
+      ? (await currentUser.getIdTokenResult()).claims.roles
       : null;
 
-    return !userClaims || !hasClaim(requireClaims, userClaims)
+    return !userClaims || !store.getters["user/hasRole"](roles, userClaims)
```
anantakrishna: Can't we just delegate dealing with the claims to the store module altogether? Why getting claims here and passing them back to the sotre getter?

frontend/src/store/index.ts
```diff
@@ -6,12 +6,17 @@ import Vue from "vue";
...
 Vue.use(Vuex);
 
 export const store = new Vuex.Store({
   state: {},
-  getters: {},
+  getters: {
+    ability() {
+      return new Ability([]);
+    }
+  },
```
anantakrishna: What is this for?

frontend/src/views/SQR/Form.vue
```diff
@@ -64,9 +64,14 @@
                 <v-btn v-if="!isCompleted" @click="saveDraft">
                   Save draft
                 </v-btn>
-                <v-btn type="submit" color="primary" class=" mx-2"
-                  >Submit</v-btn
+                <v-btn
+                  v-if="$can('update', 'SQR/Form')"
```
anantakrishna: Here semantically the action is `submit`, no?

anantakrishna: Moreover, anyone can submit if it's not Done, but only particular roles can submit Done ones. I guess this conditional rule could be handled by CASL.

frontend/src/views/TE/Task.vue
```diff
@@ -45,7 +45,7 @@
                   <v-col cols="12" sm="3" class="pt-0">
                     <v-btn
                       @click="onCancelClick"
-                      v-if="isCoordinator"
+                      v-if="$can('delete', 'TE/Task')"
```
anantakrishna: Why the action is not `cancel`?

frontend/src/views/TE/Task.vue
```diff
@@ -124,7 +124,9 @@
               </v-row>
             </v-timeline-item>
             <v-timeline-item
-              v-else-if="index === versionsCount - 1 && isCoordinator"
+              v-else-if="
+                index === versionsCount - 1 && $can('update', 'TE/Task')
```
anantakrishna: Here action is `resolve` (correct my English, the meaning is to carry out resolution).

frontend/src/views/TE/Task.vue
```diff
@@ -155,14 +157,14 @@
           </template>
         </v-timeline>
       </article>
-      <v-btn v-if="!isCoordinator" to="/te/upload">Upload</v-btn>
+      <v-btn v-if="$can('read', 'TE/Upload')" to="/te/upload">Upload</v-btn>
```
anantakrishna: I'd say `can('upload', 'TE/File`)`.

frontend/src/views/TE/Task.vue
```diff
@@ -196,19 +195,17 @@ export default class Task extends Mixins<TaskMixin, FormatTime>(
   task: any = {};
   State = State;
   state: State = State.LOADING;
-  isCoordinator = false;
+  claims!: { [key: string]: any };
```
anantakrishna: Again, I beleive the components should not deal with the claims directly.

frontend/src/views/SQR/Form.vue
DadImScared: I was using the default actions but I will switch to better named actions. I think putting this conditional rule in CASL would be difficult because data can't be easily passed into the ability. I think `!isCompleted || $can('submit', 'SQR/Form')` in the v-if might be better.

frontend/src/abilities.ts
DadImScared: If multiple roles are attached to a user I think it would be difficult to manage actions like in the link.

anantakrishna: OK, adapt the approach as needed.

frontend/src/views/SQR/Form.vue
anantakrishna: Agreed.

frontend/src/router.ts
anantakrishna: Not fixed yet.

frontend/src/views/SQR/Form.vue
```diff
@@ -264,7 +264,10 @@ export default class Form extends Vue {
     if (!allotmentStatus) {
       this.errorMessage =
         "This allotment is not valid, please contact coordinator.";
-    } else if (!isCoordinator && allotmentStatus.toLowerCase() === "done") {
+    } else if (
+      !this.hasRole("SQR.coordinator") &&
+      allotmentStatus.toLowerCase() === "done"
```
anantakrishna: Shouldn't we use `update SQR/Form done=true` ability check?

frontend/src/store/index.ts
anantakrishna: ?

frontend/src/router.ts
```diff
@@ -102,7 +102,7 @@ export const router = new Router({
               component: () => import("@/views/TE/Task.vue"),
               meta: {
                 menuItem: false,
-                auth: { requireClaims: { TE: true, coordinator: true } }
+                auth: { roles: ["TE.checker", "TE.editor", "TE.coordinator"] }
```
anantakrishna: Are you sure we can't utilize Abilities in the router? I see it as specifying the action and subject in the meta, and then checking this in the navigation guard.

For example, for this case it may be `auth: { action: 'view', subject: 'TE/Task' }`.

frontend/src/router.ts
```diff
@@ -214,11 +214,15 @@ export const checkAuth: NavigationGuard = async (to, from, next) => {
   if (guestOnly && currentUser) return next("/");
 
   if (roles) {
-    const userClaims = currentUser
+    const userRoles = currentUser
       ? (await currentUser.getIdTokenResult()).claims.roles
       : null;
```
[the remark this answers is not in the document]
DadImScared: The `checkAuth` function runs before it is populated in the store.

frontend/src/abilities.ts
```diff
@@ -1,33 +1,47 @@
...
 
 const handleSQRPermissions = (can: any) => {
   if (hasRole("SQR.coordinator")) {
-    can("manage", SQRPermissions);
+    can("manage", SQRSubjects);
     return;
   }
   if (hasRole("SQR.checker")) {
-    can("update", SQRPermissions[1]);
+    can("submit", subjects["SQR/UpdateDone"]);
```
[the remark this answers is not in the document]
DadImScared: I think the defineAbilites function can accept an object with whatever data wants to be passed and in the view the rules can be updated by passing the desired data.

frontend/src/router.ts
DadImScared: Yes I will look into refactoring it like that. Although should action be included? It is implied by having the subject.

frontend/src/store/index.ts
DadImScared: Since the ability rules were updated in the store I thought it would be better to put in the vuex store but this can be moved into it's own file and imported where rules are updated.

frontend/src/views/TE/Task.vue
```diff
@@ -45,7 +45,7 @@
                   <v-col cols="12" sm="3" class="pt-0">
                     <v-btn
                       @click="onCancelClick"
-                      v-if="$can('delete', 'TE/Task')"
+                      v-if="$can('cancel', 'TE/Task')"
```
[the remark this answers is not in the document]
DadImScared: The manage action that is assigned to the coordinator allows all actions so it is granted to the TE.coordinator role.

frontend/src/store/index.ts
anantakrishna: Maybe this should be a part of the `user` store module?

frontend/src/main.ts
```diff
@@ -55,10 +59,15 @@ async function getFirebaseConfig(): Promise<Object> {
...
-    store.commit("user/setCurrentUser", user);
+    store.dispatch("user/handleUser", user);
   });
 
-  const unsubscribe: any = firebase.auth().onAuthStateChanged(() => {
+  const unsubscribe: any = firebase.auth().onAuthStateChanged(async user => {
+    const userRoles = user
+      ? (await user.getIdTokenResult()).claims.roles
+      : null;
+    await store.dispatch("user/updateUserRoles", userRoles);
```
anantakrishna: Aren't we doning the same in the `user` store module?

frontend/src/abilities.ts
```diff
@@ -1,53 +1,77 @@
...
 };
 
-const handleSQRPermissions = (can: any) => {
-  if (hasRole("SQR.coordinator")) {
-    can("manage", SQRSubjects);
+const handleSQRPermissions = (can: any, cannot: any) => {
+  can("save", subjects.SQR.form, { isCompleted: false });
+
+  if (!hasRole(["SQR.coordinator", "SQR.checker"])) {
+    can("submit", subjects.SQR.form, { isCompleted: false });
```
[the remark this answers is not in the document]
DadImScared: The condition is if they are not `SQR.coordinator` or `SQR.checker`

frontend/src/abilities.ts
```diff
@@ -1,53 +1,77 @@
...
   }
+
+  if (hasRole("SQR.coordinator")) {
+    can(["submit", "read"], subjects.SQR.form);
+    can("allot", subjects.SQR.tasks);
+  }
+
   if (hasRole("SQR.checker")) {
-    can("submit", subjects["SQR/UpdateDone"]);
+    can(["submit", "read"], subjects.SQR.form);
```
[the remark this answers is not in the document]
DadImScared: The read permission I think might be poorly named. It's actually the permission for `allotmentStatus.toLowerCase() === "done"` and user is not a coordinator or checker. Submission is allowed regardless of `isCompleted` for the checker and coordinator and not another type of user, correct?

frontend/src/main.ts
DadImScared: This is needed for the initial page load, the one in the store is for auth state change when logging in and out.

anantakrishna: Can't we combine somehow?

frontend/src/abilities.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const hasRole = (roles: any) => {
+  return store.getters["user/hasRole"](roles);
+};
+
+const handleTEPermissions = (can: any, cannot: any) => {
+  if (hasRole("TE.coordinator")) {
+    can("manage", subjects.TE.task);
+    can("manage", subjects.TE.tasks);
```
anantakrishna: Is it implied by the framework that `manage` includes all the other actions?

frontend/src/abilities.ts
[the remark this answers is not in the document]
anantakrishna: Correct.

frontend/src/abilities.ts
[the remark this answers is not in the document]
anantakrishna: OK, I see.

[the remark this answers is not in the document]
anantakrishna: I feel it might be more explicit if all would be granted permission for `submit` action with `isCompleted: false` and elevated roles would be granted permission with `isCompleted: true`.

frontend/src/router.ts
```diff
@@ -233,26 +237,22 @@ export const checkAuth: NavigationGuard = async (to, from, next) => {
   const currentUser = firebase.auth().currentUser;
   const {
     meta: {
-      auth: { requireAuth, guestOnly, requireClaims }
+      auth: { requireAuth, guestOnly, ability }
     }
   } = restrictedRoute;
 
-  if ((requireAuth || requireClaims) && !currentUser)
+  if ((requireAuth || ability) && !currentUser)
```
anantakrishna: Consider using store to obtain `isSignedIn` property instead of getting current user from firebase directly.

frontend/src/views/SQR/Form.vue
```diff
@@ -193,13 +195,15 @@ export default class Form extends Vue {
...
-  isCoordinator = false;
   showValidationSummary = false;
-  getUserClaims!: any;
   rules = [required];
   destroyFormErrorWatch!: any;
   branch: SubmissionsBranch | null = null;
 
+  get sqrForm() {
+    return new SQRForm({ isCompleted: this.isCompleted });
+  }
```
anantakrishna: I feel it might be a bit confusing to have this property. This component is an SQR Form itself. Can't we just pass its data to the ability without introducing this property and even without SQRForm class?

DadImScared: Maybe it should be renamed to something better but this is the intended way to pass in data. https://stalniy.github.io/casl/abilities/2017/07/20/define-abilities.html#rule-conditions

frontend/src/abilities.ts
DadImScared: Yes but now I that I think about it maybe `SQR.coordinator` should have manage actions instead of specific actions for SQR phase.

frontend/src/views/SQR/Form.vue
anantakrishna: I'm not about passing the data, but about the fact that `SQR/Form` component contains `sqrForm` property which is intended only for passing to the rule. Looks weird.

More weird is that the `SQRForm` class is defined in the ability definition. Consider refactoring this. Maybe you can pass the whole form to the rule if it's easy?

frontend/src/abilities.ts
```diff
@@ -19,7 +19,7 @@ class Data {
   }
 }
 
-export class SQRForm extends Data {
+export class SQRFormAbilityData extends Data {
```
anantakrishna: Why do we ever need to define this class? Can't we just pass the data in?

frontend/src/views/SQR/Form.vue
```diff
@@ -65,7 +65,7 @@
                   Save draft
                 </v-btn>
                 <v-btn
-                  v-if="$can('submit', this.sqrForm)"
+                  v-if="$can('submit', this.abilityData)"
```
anantakrishna: You are using this data only once. I really feel it's an overhead to define the class and the property for that.

frontend/src/views/SQR/Form.vue
```diff
@@ -266,7 +266,7 @@ export default class Form extends Vue {
     } else if (
       this.$ability.cannot(
         "submit",
-        new SQRForm({
+        new SQRFormAbilityData({
           isCompleted: allotmentStatus.toLowerCase() === "done"
```
anantakrishna: Why can't we use `this.isCompleted` here as well? Actually, checking the allotment status is correct, then what is `this.isCompleted`?

frontend/src/abilities.ts
DadImScared: With default behavior the data passed in needs to have a constructor. https://stalniy.github.io/casl/abilities/2017/07/21/check-abilities.html#instance-checks

frontend/src/views/SQR/Form.vue
DadImScared: To use it in the template it has to exist on the class. Maybe a plugin that holds all the data classes would replace this need?

frontend/src/views/SQR/Form.vue
DadImScared: `this.isCompleted` returns true when data is loaded from branch `completed` or `migrated`.

anantakrishna: I feel that the proper way is to check the allotment status rather than the source branch of the form. Is it possible for both places?

And it might be better to rename the `isCompleted` property to something like `isLoadedFromCompletedBranch` for clarity.

anantakrishna: Moreover! I have just realized that it is absolutely improper to rely on the source branch, as the form may be completed (submitted) and still it is available for re-submission by the author until it is marked as Done by the checker. So we should rely on the allotment status exclusively.

frontend/src/views/SQR/Form.vue
```diff
@@ -64,9 +64,19 @@
...
                   Save draft
                 </v-btn>
-                <v-btn type="submit" color="primary" class=" mx-2"
-                  >Submit</v-btn
+                <v-btn
+                  v-if="
+                    $can('submit', {
+                      modelName: $subjects.SQR.form,
+                      isCompleted
+                    })
```
anantakrishna: Allotment status should be considered here, not the source branch.

frontend/src/abilities.ts
```diff
@@ -0,0 +1,59 @@
...
+    can("view", subjects.TE.tasks);
+  }
+  if (hasRole("TE.editor")) {
+    can(["upload", "view"], subjects.TE.task);
+    can("view", subjects.TE.myTasks);
+  }
+};
+
+const handleSQRPermissions = (can: any, cannot: any) => {
+  can("submit", subjects.SQR.form, { isCompleted: false });
```
anantakrishna:
```suggestion
  can("submit", subjects.SQR.form, { isMarkedDone: false });
```

frontend/src/views/SQR/Form.vue
```diff
@@ -61,12 +61,22 @@
...
                 </v-btn>
-                <v-btn type="submit" color="primary" class=" mx-2"
-                  >Submit</v-btn
+                <v-btn
+                  v-if="
+                    $can('submit', {
+                      modelName: $subjects.SQR.form,
+                      isMarkedDone
+                    })
+                  "
```
anantakrishna: Add an alternative message for those who don't have access to Submit. Message: “Form is marked as done. Please contact the coordinator.”

DadImScared: When the allotment is validated if they have no access to submit this is what's shown 
image

anantakrishna: Oh I see. Let's refactor it a bit (I'll add to the card as well).

If the form is marked as Done then:

- The form should still load.
- There should be a warning alert in the top “The submission is finalized.”
- All the controls should be read only and there should be no Submit button — for those who cannot update finalized submissions.

PR #324, opened 2020-01-09
Indicate that the TE task is based on the restored files

DadImScared:
https://trello.com/c/tcf8xk0d/295-indicate-that-the-te-task-is-based-on-the-restored-files

### DIFF
--- a/frontend/src/components/TE/TaskDefinition.vue
+++ b/frontend/src/components/TE/TaskDefinition.vue
@@ -3,14 +3,15 @@
     <template v-for="(chunk, index) in item.chunks">
       <v-row class="wrap pb-1" :key="`${item['key']}-${index}-layout`">
         <v-col v-bind="layout.link">
-          <div :style="{ display: 'flex', flexWrap: 'wrap' }">
-            <v-chip v-if="item.isRestored" class="mr-1" color="primary">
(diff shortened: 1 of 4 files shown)


frontend/src/components/TE/TaskDefinition.vue
```diff
@@ -3,11 +3,15 @@
...
+          <v-badge v-if="item.isRestored" color="#2196F3">
+            <template v-slot:badge>SEd</template>
             <a :href="getLink(chunk.fileName)">
               {{ chunk.fileName }}
             </a>
-          </div>
+          </v-badge>
+          <a v-else :href="getLink(chunk.fileName)">
+            {{ chunk.fileName }}
+          </a>
```
anantakrishna: Is it possible to avoid duplication of `a` tag?

Also, as I've mentioned in the card just now, the badge should be added to the task ID, not the source file name.

DadImScared: Without something in the badge slot a small blue circle is present.

anantakrishna: Then let it be always badge, but in case `restored is false` put `Non-SEd` into the badge.

PR #325, opened 2020-01-18
Check validity of the TE task before uploading the file

DadImScared:
https://trello.com/c/zGhF9XuA/270-check-validity-of-the-te-task-before-uploading-the-file


frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+        spark.append(event.target.result);
+        resolve(btoa(spark.end(true)));
+      };
+      fileReader.readAsArrayBuffer(file);
+    });
+  }
+
+  async validateFileHash(file: File, task: any) {
+    const fileHash = await this.getFileHash(file);
+    if (!task.versions) return;
```
anantakrishna: No need to calculate the hash if there is no version.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    for (const version of Object.values<any>(task.versions)) {
+      const ref = this.uploadsBucket.ref().child(`/${version.uploadPath}`);
```
anantakrishna: Is it necessary to prefix the path with `/`?

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+      const metadata = await ref.getMetadata().catch(e => "error");
+      if (metadata === "error") continue;
+      if (fileHash === metadata.md5Hash) {
+        throw new Error("You had uploaded the same file earlier.");
```
anantakrishna: Add the version number and the date/time of the version.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+      "resolution.timestamp"
+    );
+    if (
+      lastTaskResolutionTimestamp &&
+      lastTaskResolutionTimestamp > file.lastModified
+    ) {
+      throw new FileUploadError({
+        props: { taskId },
+        component: "file-time-error"
+      });
```
anantakrishna: Why can't we throw general `Error` here? What makes this check different? It complicates the logic, makes it non-uniform.

frontend/src/views/TE/Upload.vue
```diff
@@ -198,18 +273,17 @@ export default class Upload extends Vue {
...
-        this.emitFileError(file, e.message);
-      }
-    });
+    for (const file of files) {
+      this.updateFileFields(file, {});
+      this.validateUpload(file)
+        .then(() => this.handleFile(file, timestamp))
+        .catch(e =>
+          this.emitFileError(file, e instanceof FileUploadError ? e : e.message)
+        );
```
anantakrishna: Is there any particular reason why you changed to chaining `then` and `catch` instead of `try ... catch`? Perhaps because the validation became async... Can't we just `await` it?

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+      const ref = this.uploadsBucket.ref().child(`/${version.uploadPath}`);
+      const metadata = await ref.getMetadata().catch(e => "error");
+      if (metadata === "error") continue;
+      if (fileHash === metadata.md5Hash) {
+        throw new Error("You had uploaded the same file earlier.");
+      }
+    }
+  }
+
+  async validateUpload(file: File) {
```
anantakrishna:
```suggestion
  async validateFile(file: File) {
```

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    validateFlacFile(file);
```
anantakrishna:
```suggestion
    validateContentType(file);
```

anantakrishna: Moreover, I see that this function is validating both content type and file name.

I think due to our current refactoring, we can

- File name can be validated directly in this method as a step. Task ID regexp can be imported.
- Content type can be validated directly here as a step. It's just a line of code, and it doesn't worth having a separate function and tests for it.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    const taskId = getTaskId(file.name);
+    const task = (
+      await firebase
+        .database()
+        .ref(`/TE/tasks/${taskId}`)
+        .once("value")
+    ).val();
+    if (this.currentUser.email !== _.get(task, "assignee.emailAddress")) {
+      throw new Error(`The task ${taskId} is not assigned to you.`);
+    }
```
anantakrishna: According to the database rules, the user will be able to read the task only if she is an assignee, so this check is redundant. Instead, you should validate that you have not got access violation.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    if (task.status === "Done") {
+      throw new Error(`The task ${taskId} is marked as Done.`);
```
anantakrishna:
```suggestion
      throw new Error(`The task ${taskId} is marked as Done. Uploads are not allowed.`);
```

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    }
+    const lastTaskResolutionTimestamp = _.get(
```
anantakrishna:
```suggestion
    const lastVersionResolutionTimestamp = _.get(
```

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
...
+    if (
+      lastTaskResolutionTimestamp &&
+      lastTaskResolutionTimestamp > file.lastModified
+    ) {
+      throw new FileUploadError({
+        props: { taskId },
+        component: "file-time-error"
+      });
+    }
+    await this.validateFileHash(file, task);
```
anantakrishna: This method is not too big, consider moving its contents here for clarity. I mean, all the validation steps will be directly in this `validateFile` method.

frontend/package.json
```diff
@@ -12,12 +12,14 @@
     "test:unit": "npm run test"
   },
   "dependencies": {
+    "@types/spark-md5": "^3.0.2",
```
anantakrishna: Types usually are `devDependencies`.

frontend/src/views/TE/Upload.vue
DadImScared: What would be an access violation? Someone who is not an editor or coordinator or is not assigned to the task?

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +209,62 @@ export default class Upload extends Vue {
     }
   }
 
+  getFileHash(file: File) {
+    return new Promise(resolve => {
```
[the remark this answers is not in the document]
DadImScared: Yes we have to turn `fileReader.onload` event into a promise so we can async/await it when getting the hash.

frontend/src/views/TE/Upload.vue
anantakrishna: I'm not sure I understand the question.
Currently according to the database rules, only assignee, coordinator and checker can access a task.

DadImScared: Sorry, my question was poorly phrased. I understand now though. An editor can also upload files should they be allowed access too?

frontend/src/views/TE/Upload.vue
DadImScared: Using `await` would prevent the next file validation until the one before it is validated.

frontend/src/views/TE/Upload.vue
DadImScared: General `Error` only accepts a string.

frontend/src/views/TE/Upload.vue
anantakrishna: Assignee = editor, who is actually assigned to the current task.

frontend/src/views/TE/Upload.vue
anantakrishna: I don't see a problem. Passing error string is enough, like in the other cases.

frontend/src/views/TE/Upload.vue
anantakrishna: I see, good point.

frontend/src/views/TE/Upload.vue
DadImScared: This error is required to display a link to the task but can't create a `router-link` in a string.

anantakrishna: I see now. Let's proceed without a task link as of now, I've removed this requirement from the card.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +194,65 @@ export default class Upload extends Vue {
...
+    }
+    if (!task.versions) return;
+    const fileHash = await this.getFileHash(file);
+    for (const [versionId, version] of Object.entries<any>(task.versions)) {
+      const ref = this.uploadsBucket.ref().child(version.uploadPath);
+      const metadata = await ref.getMetadata().catch(e => "error");
+      if (metadata === "error") continue;
+      if (fileHash === metadata.md5Hash) {
+        throw new Error(
+          `You had uploaded the same file earlier. Version: ${versionId} on ${moment(
```
anantakrishna: `versionId` is non-human-readable key, show the version number (1,2,3,…) here.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +194,65 @@ export default class Upload extends Vue {
...
+      throw new Error("The file name is not a correct task ID");
+    }
+    const taskId = getTaskId(file.name);
+    const task = (
+      await firebase
+        .database()
+        .ref(`/TE/tasks/${taskId}`)
+        .once("value")
+    ).val();
+    if (task.status === "Done") {
```
anantakrishna: What if the task does not exist or not accessible to the user? I guess you missed “The task {{taskId}} is not assigned to you.” check here.

frontend/src/views/TE/Upload.vue
```diff
@@ -190,6 +194,65 @@ export default class Upload extends Vue {
...
+      this.getLastVersion(task),
+      "resolution.timestamp"
+    );
+    if (
+      lastVersionResolutionTimestamp &&
+      lastVersionResolutionTimestamp > file.lastModified
+    ) {
+      throw new Error("File is older than the latest feedback on the task.");
+    }
+    if (!task.versions) return;
```
anantakrishna: By the way, you can move this check above the previous check.

frontend/src/views/TE/Upload.vue
DadImScared: When the task is not accessible firebase will throw a permission denied error which is displayed. Should this be detected and replaced with a different message?

anantakrishna: Depends on the message that Firebase throws. Can you show it to me?

PR #330, opened 2020-02-09
TE upload optimization

DadImScared:
https://trello.com/c/8e43n642/320-te-upload-optimization


frontend/src/views/TE/Upload.vue
```diff
@@ -44,75 +44,59 @@
...
 import moment from "moment";
+import Promise from "bluebird";
 import firebase from "firebase/app";
 import "firebase/database";
 import "firebase/storage";
 import BaseTaskMixin from "@/components/TE/BaseTaskMixin";
+import UploadFileList from "@/components/TE/UploadFileList.vue";
+
+import { getTaskId, getProjectDomain } from "@/utility";
+import UploadFiles from "@/components/TE/UploadFilesList.vue";
```
anantakrishna: It is confusing to have two similar components: `UploadFilesList` and `UploadFileList`.

I am not sure it's worth extracting `UploadFilesList` into a separate component. Let's keep it within `Upload` component.

frontend/src/components/TE/UploadFilesList.vue
```diff
@@ -0,0 +1,74 @@
...
+@Component({
+  name: "UploadFilesList",
+  components: {
+    FileList: UploadFileList
+  }
+})
+export default class UploadFilesList extends Vue {
+  @Prop({ default: () => [] }) uploadingFiles!: any[];
+  @Prop({ default: () => [] }) queuedFiles!: File[];
+  @Prop({ default: () => [] }) completedFiles!: File[];
```
anantakrishna: I expected that we'd have a single queue/list. I supposed that we can sort the list by the status+timestamp and thus get the proper order of items.

Please take a look at https://github.com/sindresorhus/p-queue and https://github.com/sindresorhus/promise-fun, maybe you will be inspired how to implement queues more elegantly.

PR #336, opened 2020-02-15
Add ability to select Task ID text on the allotment form

DadImScared:
https://trello.com/c/MlQRyyyq/338-add-ability-to-select-task-id-text-on-the-allotment-form

### DIFF
--- a/frontend/src/styles/main.css
+++ b/frontend/src/styles/main.css
@@ -1,4 +1,4 @@
 
-.user-select-text {
+.v-input--checkbox .v-label {
     user-select: text;
 }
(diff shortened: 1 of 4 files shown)


frontend/src/styles/main.css
```diff
@@ -0,0 +1,4 @@
+
+.user-select-text {
+    user-select: text;
+}
```
anantakrishna: I was supposing we're going to override standard Vuetify styles to make the labels selectable.

DadImScared: Should this be done with all labels or just per page as needed?

anantakrishna: Globally, for all labels.

PR #342, opened 2020-02-24
Display circular progess in WIP/Given stats on the allotment form

DadImScared:
https://trello.com/c/05kUAROZ/342-display-wip-given-stats-on-the-allotment-form-once-assignee-is-chosen

### DIFF
--- a/frontend/src/views/TE/Allotment.vue
+++ b/frontend/src/views/TE/Allotment.vue
@@ -19,10 +19,30 @@
             :style="{ visibility: allotment.assignee ? 'visible' : 'hidden' }"
           >
             <v-chip :color="getTaskStyle({ status: 'Given' }).backgroundColor">
-              Given {{ assigneeTasksStats.Given }}
+              Given
+              <v-avatar right>
+                <v-progress-circular
+                  size="24"
+                  indeterminate
+                  v-if="isLoadingAssigneeTasks"
+                ></v-progress-circular>
+                <template v-else>
+                  {{ assigneeTasksStats.Given }}
+                </template>
+              </v-avatar>
             </v-chip>
             <v-chip :color="getTaskStyle({ status: 'WIP' }).backgroundColor">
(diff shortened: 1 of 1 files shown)


frontend/src/views/TE/Allotment.vue
```diff
@@ -19,10 +19,30 @@
             :style="{ visibility: allotment.assignee ? 'visible' : 'hidden' }"
           >
             <v-chip :color="getTaskStyle({ status: 'Given' }).backgroundColor">
-              Given {{ assigneeTasksStats.Given }}
+              Given
+              <v-avatar right>
+                <v-progress-circular
+                  size="24"
+                  indeterminate
+                  v-if="isLoadingAssigneeTasks"
```
anantakrishna: Can we rely on `assigneeTasksStats.Given` to be `undefined` for this? I think it might be better than introducing another property.

DadImScared: Currently there are default values of 0. If this is changed, after the tasks are loaded if the user has not been assigned any tasks yet it will always show undefined.

anantakrishna: I don't think `0` is a good default. As in reality we don't know the count, which in fact could be `0`. Thus, the two states are merged: unknown and 0.

So please refactor the code to have the values `undefined` by default (or `null` if it makes more sense) and replace it with `0` once the tasks are loaded and counted.

DadImScared: The default values are in the reduce function [here](https://github.com/PureBhaktiArchive/audioseva/blob/49726fde78789e289a831a802f834b58231b4361/frontend/src/views/TE/Allotment.vue#L248). This prevents extra code required in the reduce callback function and in the template. With or without the default values in the reduce function we still have to have the loading variable for tasks. Should it be refactored still?

anantakrishna: OK, not now.

PR #351, opened 2020-03-11
Apply user roles changes instantly

DadImScared:
https://trello.com/c/7SlMr7Us/316-apply-user-roles-changes-instantly


database.rules.bolt
```diff
@@ -181,3 +181,9 @@ path /quotes {
     read() { true }
     path /{push_id}/text is String {}
 }
+
+path /metadata/{user_id}/refreshTime {
```
anantakrishna: Please use the existing path `/users/{uid}/`.
```suggestion
path /users/{uid}/refreshTime {
```

frontend/src/main.ts
```diff
@@ -58,14 +58,31 @@ async function getFirebaseConfig(): Promise<Object> {
...
 getFirebaseConfig().then(config => {
   firebase.initializeApp(config);
   firebase.auth().onAuthStateChanged(user => {
-    store.dispatch("user/handleUser", user);
+    store.dispatch("user/handleUser", { user });
+    if (callback && metadataRef) {
+      metadataRef.off("value", callback);
+    }
+    if (user) {
+      metadataRef = firebase.database().ref(`metadata/${user.uid}/refreshTime`);
```
anantakrishna: Why `main.ts` is doing this? Perhaps `user` store module would fit more.

frontend/src/store/modules/user.ts
```diff
@@ -59,13 +59,16 @@ export default {
     },
     async handleUser(
```
anantakrishna: Now it's not that clear what does this method is intended for. Please consider renaming.

frontend/src/main.ts
```diff
@@ -60,8 +60,9 @@ async function getFirebaseConfig(): Promise<Object> {
 
 getFirebaseConfig().then(config => {
   firebase.initializeApp(config);
-  firebase.auth().onAuthStateChanged(user => {
-    store.dispatch("user/handleUser", user);
+  firebase.auth().onAuthStateChanged(async user => {
+    await store.dispatch("user/handleUser", user);
+    store.dispatch("user/handleMetaData");
```
anantakrishna: Why can't we move this code altogether into the `user` store module? I mean hooking on `onAuthStateChanged`. From here we can call an action of the `user` module indicating that the config is obtained, and that the module can use Firebase functionality. It may be even more general store event.

frontend/src/main.ts
```diff
@@ -60,8 +60,9 @@ async function getFirebaseConfig(): Promise<Object> {
...
   });
 
   const unsubscribe: any = firebase.auth().onAuthStateChanged(async user => {
```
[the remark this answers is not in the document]
anantakrishna: Perhaps to wait for the first user info to come, to delay the `Vue` app construction till then. Please consider, whether we need to wait now, when we already have instant roles propagation throughout the UI.

frontend/src/store/modules/user.ts
```diff
@@ -62,21 +65,39 @@ export default {
...
+    },
+    async getUserRoles(
+      { state }: ActionContext<any, any>,
+      refreshToken = false
+    ) {
+      if (!state.currentUser) return null;
+      return (await state.currentUser.getIdTokenResult(refreshToken)).claims
+        .roles;
+    },
+    handleMetaData({ state, dispatch }: ActionContext<any, any>) {
```
anantakrishna: The name is confusing. At first I thought that it actually handles the metadata change, similar to `handleUser`.

Can't we move all this code into the `setCurrentUser` mutation? Because we need to re-hook only when the user changes. I see no async code here, so why not to move?

frontend/src/store/modules/user.ts
```diff
@@ -10,6 +10,9 @@ import _ from "lodash";
 import { Ability } from "@casl/ability";
 import { defineAbilities } from "@/abilities";
 
+let callback: ((snapshot: firebase.database.DataSnapshot) => void) | null;
+let metadataRef: firebase.database.Reference | null;
```
anantakrishna: I don't like that we have these global variables. Can't we make them module properties?

frontend/src/store/modules/user.ts
```diff
@@ -62,21 +65,39 @@ export default {
...
+    ) {
+      if (!state.currentUser) return null;
+      return (await state.currentUser.getIdTokenResult(refreshToken)).claims
+        .roles;
+    },
+    handleMetaData({ state, dispatch }: ActionContext<any, any>) {
+      if (callback && metadataRef) {
+        metadataRef.off("value", callback);
+      }
+      if (state.currentUser) {
```
anantakrishna: When the rest of the method is inside `if`, better to negate and return.

frontend/src/store/modules/user.ts
```diff
@@ -62,21 +65,39 @@ export default {
...
+        callback = async () => {
+          await dispatch("onMetaDataChange");
+        };
+        metadataRef = firebase
+          .database()
+          .ref(`users/${state.currentUser.uid}/refreshTime`);
+        metadataRef.on("value", callback);
```
anantakrishna: I think we should track the last time when we refreshed the roles and compare it to the new value in the callback. Because “This event will trigger once with the initial data stored at this location, and then trigger again each time the data changes.” (from https://firebase.google.com/docs/reference/js/firebase.database.Reference#value-event). Thus, we will get one callback after hooking even if the time was not changed.

OR

We can use this feature to update the roles after user change. Thus, we don't need to dispatch `updateUserRoles` in `handleUser`.

Consider which approach is better. I like the latter.

frontend/src/main.ts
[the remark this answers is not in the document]
DadImScared: We need to populate the roles before the app loads so that routing has access to them to determine if a route is allowed.

frontend/src/store/modules/user.ts
```diff
@@ -62,21 +65,39 @@ export default {
...
       { commit, getters }: ActionContext<any, any>,
       roles: { [key: string]: any } | null
     ) {
       commit("setUserRoles", roles);
       getters.ability.update(defineAbilities());
+    },
+    async getUserRoles(
+      { state }: ActionContext<any, any>,
+      refreshToken = false
+    ) {
```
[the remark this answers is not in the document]
DadImScared: We need to access the ability through getters to update it and we can't access getters in a mutation.

frontend/src/main.ts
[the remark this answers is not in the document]
anantakrishna: Yes, I remember that. However look, later on the roles can be changed, and the app adapts instantly. The routing adapts. Maybe we can create the app here before the roles are fetched, and when they are fetched, the current route will adapt and all the UI will adapt.

frontend/src/store/modules/user.ts
```diff
@@ -14,7 +14,10 @@ export default {
   namespaced: true,
   state: {
     currentUser: null,
-    roles: null
+    roles: null,
+    metadataCallback: null,
+    metadataRef: null,
```
anantakrishna: I'm sure there is no need keeping this as a property, as we can construct the ref using `currentUser.uid`.

anantakrishna: At least it can become a getter.

frontend/src/store/modules/user.ts
```diff
@@ -51,32 +54,69 @@ export default {
...
+      { commit, getters, state }: ActionContext<any, any>,
+      { forceRefresh = false } = {}
     ) {
+      const roles = state.currentUser
+        ? (await state.currentUser.getIdTokenResult(forceRefresh)).claims.roles
+        : null;
       commit("setUserRoles", roles);
       getters.ability.update(defineAbilities());
+    },
+    handleMetadata({ state, dispatch, commit }: ActionContext<any, any>) {
```
anantakrishna: This action is used only from `onAuthStateChanged`. And it actually should be called only when the user is changed. So, I propose moving the body of this action directly into the `onAuthStateChanged` action.

frontend/src/main.ts
```diff
@@ -60,12 +60,12 @@ async function getFirebaseConfig(): Promise<Object> {
...
   firebase.initializeApp(config);
-  firebase.auth().onAuthStateChanged(user => {
-    store.dispatch("user/handleUser", user);
+  firebase.auth().onAuthStateChanged(async user => {
+    store.dispatch("user/onAuthStateChanged", user);
   });
 
   const unsubscribe: any = firebase.auth().onAuthStateChanged(async user => {
-    await store.dispatch("user/handleUser", user);
+    await store.dispatch("user/handleInitialUserLoad", user);
```
anantakrishna: I guess we don't need to dispatch this action from this callback. The other callback above is already dispatching an action to the store to set the current user. Please think of another way to wait for the roles to be ready here.

frontend/src/store/modules/user.ts
```diff
@@ -51,32 +54,69 @@ export default {
...
+      if (state.metadataCallback && state.metadataRef) {
+        commit("setTimestamp", null);
+        state.metadataRef.off("value", state.metadataCallback);
+      }
+      if (!state.currentUser) {
+        return;
+      }
```
anantakrishna: Perhaps this means that the user has signed out. Don't we need to reset the abilities in this case?

frontend/src/main.ts
DadImScared: I have found that the roles do exist in state but the getter `hasRole` is not working as intended. With this knowledge I should have a better solution to this problem.

frontend/src/store/modules/user.ts
```diff
@@ -52,31 +60,55 @@ export default {
...
       }
+      if (!state.currentUser) {
+        commit("setMetadataCallback", null);
+        commit("setTimestamp", null);
+        await dispatch("updateUserRoles");
+        return;
+      }
+      const callback = () => {
+        if (!state.timestamp) {
+          commit("setTimestamp", +new Date());
```
anantakrishna, line 95: I'm afraid the logic here is not correct. We should save the `refreshTime` that we got from the database. By the way, it's a good candidate for the state property name: `metadataRefreshTime`.

Also, it might be useful to compare the refresh time from the database against the local last refresh time. And avoid updating the roles if the local refresh time is higher.

frontend/src/store/modules/user.ts
```diff
@@ -52,31 +60,55 @@ export default {
...
-      if (user) {
-        await dispatch(
-          "updateUserRoles",
-          (await user.getIdTokenResult()).claims.roles
-        );
-      } else {
-        await dispatch("updateUserRoles", null);
+
+      if (state.metadataCallback && getters.metadataRef) {
+        commit("setTimestamp", null);
```
anantakrishna, line 84: I think setting it to null is not a part of this piece of code. Here you only remove the callback. The timestamp (refresh time) should be set to current time along with updating the roles after the user signs in.

